PR #38, opened 2023-03-21
Refactor/decouple pretty table

franjpr:
Decouple prettytable dependency from the base_estimator class


cryptographic_estimators/base_estimator.py
```diff
@@ -258,64 +258,7 @@ def estimate(self, **kwargs):
...
-        """
-        for i in self.estimates.keys():
-            row = [self.estimates[i][tbl.title][k] for k in tbl.field_names]
-            row = [round_or_truncate(i, truncate, precision)
-                   if i in RR else i for i in row]
-            tbl.add_row(row)
-        return tbl
-
-    def table(self, show_quantum_complexity=0, show_tilde_o_time=0, show_all_parameters=0, precision=1, truncate=0):
+    def table(self, show_quantum_complexity=false, show_tilde_o_time=false, show_all_parameters=false, precision=1, truncate=false):
```
Memphisd, line 261: python false is written with capital F so False

cryptographic_estimators/base_estimator.py
```diff
@@ -328,37 +271,12 @@ def table(self, show_quantum_complexity=0, show_tilde_o_time=0, show_all_paramet
...
-            if j == BASE_QUANTUMO:
-                tbl._min_width = {BASE_TIME: len(BASE_QUANTUMO)}
-
-        tbl_join = concat_all_tables(tables)
-
-        print(tbl_join)
+        renderer = EstimationRenderer(
+            show_quantum_complexity, show_tilde_o_time, show_all_parameters, precision, truncate
+        )
+        renderer.as_table(self.estimate())
```
Memphisd, line 279: it should estimate first and if the return of estimate is an empty dictionary it should throw an error "No algorithm associated with this estimator or applicable to this problem instance"

Memphisd: if not empty it can be passed to the renderer

cryptographic_estimators/estimation_renderer.py
```diff
@@ -0,0 +1,115 @@
...
+        self._show_all_parameters = show_all_parameters
+        self._precision = precision
+        self._truncate = truncate
+
+    def as_table(self, estimation_result: dict) -> None:
+        """
+        Prints the given estimation dictionary as a table
+        """
+        estimation = deepcopy(estimation_result)
+        if estimation is None:
```
Memphisd, line 31: as said I think this check should be outside the printing logic. IF it remains here the error thrown would have to be something like "Empty dictionary inputed for printing"

cryptographic_estimators/estimation_renderer.py
```diff
@@ -0,0 +1,115 @@
...
+            print("No algorithms associated with this estimator.")
```
Memphisd, line 32: I suggest to change the print to an exception (in the base_estimator file)

Memphisd (changes requested):
Check for empty estimates should be done in base estimator class

cryptographic_estimators/base_estimator.py
franjpr: Ok thanks, I will fix this

franjpr:
I'm going to open a new PR, because this thas weird conflicts while merging develop

PR #40, opened 2023-03-22
refactor/add_pkp_input_dic

Javierverbel:


cryptographic_estimators/PKEstimator/pk_problem.py
```diff
@@ -29,9 +29,9 @@ class PKProblem(BaseProblem):
     - ``n`` -- columns of the matrix
     - ``m`` -- rows of the matrix
     - ``q`` -- size of the field
-    - ``ell`` -- rows of the matrix whose permutation should lie in the kernel
-    - ``use_parity_row`` -- enables trick of appending extra (all one) row to the matrix, i.e., m -> m+1 (default:False)
-    - ``nsolutions`` -- number of (expected) solutions of the problem in logarithmic scale
+    - ``ell`` -- number of rows of the matrix whose permutation should lie in the kernel (default: 1)
+    - ``use_parity_row`` -- enables trick of appending extra (all one) row to the matrix, i.e., m -> m+1 (default: false)
+    - ``nsolutions`` -- number of solutions of the problem in logarithmic scale (default: expected_number_solutions)
```
Memphisd, line 34: I just realized, that this is not entirely true, as the default is the maximum between the  expected number and 1 (or 0 in log scale)

cryptographic_estimators/MQEstimator/mq_problem.py
```diff
@@ -36,7 +36,7 @@ class MQProblem(BaseProblem):
     - ``n`` -- number of variables
     - ``m`` -- number of polynomials
     - ``q`` -- order of the finite field (default: None)
-    - ``nsolutions`` --  number of (expected) solutions of the problem in logarithmic scale
+    - ``nsolutions`` --  number of solutions of the problem in logarithmic scale (default: expected_number_solutions)
```
Memphisd, line 39: same here

Memphisd (changes requested):

input_dictionary.json
```diff
@@ -461,6 +459,130 @@
...
+				"display_label": "Number of rows",
+				"placeholder": "Insert parameter",
+				"tooltip": "Number of rows of the input matrix"
+			},
+           {
+				"id": "q",
+				"type": "number",
+				"display_label": "Field size",
+				"placeholder": "Insert parameter",
+                "tooltip": "An integer of the form p^x for some integer x, where p is prime number.  This value indicates the number of elements in the underlying field"
```
Memphisd, line 488: actually for PKP I think the q should be prime, defined over extension fields it might be less secure

Memphisd: I mean the algorithms still work in that case, but it might give the impression we claim that in this case those algorithms still yield a good estimate for the general hardness (meaning they are still the best, which is likely not the case)

input_dictionary.json
```diff
@@ -461,6 +459,130 @@
...
+          {
+				"id": "ell",
+				"type": "number",
+				"display_label": "No. of rows in the kernel",
+                "placeholder": "Insert value",
+                "default_value": 1,
+                "tooltip": "Number of rows of the matrix whose permutation should lie in the kernel"
+          },
+          {
+				"id": "use_parity_row",
```
Memphisd, line 500: this is not a problem parameter, is it?

input_dictionary.json
```diff
@@ -461,6 +459,130 @@
...
+              "direction": "column",
+              "display_label": "Included algorithms",
+              "tooltip": "Algorithms to include for optimization",
+              "default_value": [],
+              "excluded_algorithms": [],
+              "options": [],
+              "dependencies": []
+        },
+        {
+              "id": "include_tildeo",
```
Memphisd, line 544: even though no tilde o is implemented we want to show the switch in the frontend?

Memphisd (changes requested):

input_dictionary.json
Javierverbel, line 500: @Andre this shows as problem parameter in the pk_problem.py

input_dictionary.json
Javierverbel: Now it says in the tooltip it is a prime number, is it enough?

input_dictionary.json
Javierverbel: We shouldn't, it might cause confusion.

Memphisd (dismissed):

Memphisd (approved):

PR #52, opened 2023-03-29
Feat/test mq

Javierverbel:


Javierverbel:
In this branch add the tests of the MQ estimator. 

Additionally it reverts some opimizations of the crossbred algorithm. Such a optimization sometimes yield estimates 1 bit smaller then the previous MQ.

Memphisd (approved):
LGTM, all tests passing.

PR #53, opened 2023-03-29
Refactor/some polishing

Memphisd:
Reviewed mainly MQ estimator and changed to using the function for setting parameters. + Some overall polishing.


cryptographic_estimators/main.py
FloydZ, line 1: should be removed?

FloydZ (commented):
I think main.py should be removed

FloydZ (approved):

PR #71, opened 2023-08-24
refactor: memory access cost estimation

Javierverbel:
### Description
- Refactor the estimation of the memory access cost.

### Review process
- Make sure the estimation of the memory access cost is correct.


Memphisd (approved):
Just to add some more context for the documentation of this change:
Javier observed, that the memory access cost actually is the cost in bits for memory access per basic operation. Therefore, T*f(M) where f(M) is the memory access cost is already in bit unit. The total complexity is therefore T*f(M) + T * #bit_operations_per_basic_operation and not
 T*f(M) * #bit_operations_per_basic_operation

PR #86, opened 2023-10-12
fix: display tilde o in table when complexity type 1

Memphisd:
### Description
When complexity type is set to one the table function now always includes the tilde o estimate.


FloydZ (approved):

Memphisd:
re-created this PR, since it was accidentally requested to be merged into main initially. Check for merge source not re-triggered after changing to develop (also a PR incoming to fix that triggering)

PR #87, opened 2023-10-13
Fix/display tilde o v2

FloydZ:
### Description
- When complexity type is set to one the table function now always includes the tilde o estimate.


Memphisd (approved):
looks good

PR #94, opened 2023-10-25
Fix/time_complexity_and_speed_lokshtanov

Javierverbel:
### Description
Faster Lokshtanov complexity estimation.
Remove unecessary factor in the time complexity formula.


cryptographic_estimators/MQEstimator/MQAlgorithms/lokshtanov.py
```diff
@@ -120,7 +112,7 @@ def _compute_time_complexity(self, parameters: dict):
             sage: from cryptographic_estimators.MQEstimator.mq_problem import MQProblem
             sage: E = Lokshtanov(MQProblem(n=10, m=12, q=9), bit_complexities=False)
             sage: E.time_complexity(delta=2/10)
-            214.16804105519708
+            210.99786719362038
```
FloydZ, line 115: where is this improvement coming from, if the only thing you did, was to move the function `_C` into the computation loop of `_compute_time_complexity`?

FloydZ (commented):
Only a single question about the improvement

cryptographic_estimators/MQEstimator/MQAlgorithms/lokshtanov.py
Javierverbel: I had an additional factor in the time complexity formula. I double check it shouldn't be in there.

Javierverbel: Notice commit `@Javierverbel
[Fix: unnecesary factor of n in time complexity](https://github.com/Crypto-TII/CryptographicEstimators/pull/94/commits/994c862d846dd917610c824a764895687641ad8a)`

PR #95, opened 2023-10-25
Fix/memory complexity boolean solve fxl

Javierverbel:
### Description
- Fix complexity memory complexity of BooleanSolve and FXL algorithms


cryptographic_estimators/MQEstimator/mq_estimator.py
```diff
@@ -52,16 +52,16 @@ class MQEstimator(BaseEstimator):
...
             | BooleanSolveFXL  | 46.3 |   16.1 |
-            | Crossbred        | 44.7 |   30.6 |
+            | Crossbred        | 39.8 |   37.9 |
             | DinurFirst       | 57.7 |   37.9 |
             | DinurSecond      | 42.8 |   33.6 |
             | ExhaustiveSearch | 44.4 |   16.1 |
-            | F5               | 86.2 |   57.0 |
-            | HybridF5         | 49.8 |   16.1 |
+            | F5               | 62.9 |   57.0 |
+            | HybridF5         | 48.6 |   16.1 |
```
FloydZ, line 60: These are dramatic improvements! But why is crossbred improving and not `BooleanSolveFXL`, if the onlye changes were made to the latter?
Anyway this change should be documented somewhere, or?

FloydZ (commented):
Only a single comment about the time improvements.

cryptographic_estimators/MQEstimator/mq_estimator.py
Javierverbel: The change comes from a change in the test. Notice now it is using `w = 2`.

FloydZ: ah ok.

PR #96, opened 2023-10-30
Feature/implement minrank estimator

Javierverbel:
### Description
- First version of MinRank estimator. Only includes one algorithms

### Review process
- Review base mr clases.
- Review support_minors module.


cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
```diff
@@ -0,0 +1,261 @@
...
+            sage: from cryptographic_estimators.MREstimator.MRAlgorithms.support_minors import SupportMinors
+            sage: from cryptographic_estimators.MREstimator.mr_problem import MRProblem
+            sage: SM = SupportMinors(MRProblem(q=7, m=9, n=10, k=15, r=4))
+            sage: SM.nprime()
+            'strassen'
+        """
+        return self._get_optimal_parameter(MR_VARIANT)
+
+
+    def _expected_dimension_of_support_minors_equations(self,q, m, n, K, r, b):
```
FloydZ, line 125: I can imagine that this functions is quite slow. Would it be better to translate all `* binomial(x, y)` to `+ log2(...)`?

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
```diff
@@ -0,0 +1,261 @@
...
+        variant = MR_STRASSEN
+        keys = [i for i in indices]
+        stop = False
+        while not stop:
+            if not self._are_parameters_invalid(indices):
+                aux = indices.copy()
+                aux.update({MR_VARIANT: variant})
+                yield aux
+            indices[next(iter(indices))] += 1
+            for i in range(len(keys)):
```
FloydZ, line 184: Thats seems hacky: As far as I understand you need to override to original `_valid_choices` to be able to change the matrix algorithm from Strassen to Block_Wiedemann. I think it would far easier to define these two types as an enumeration:
```python
from enum import Enum

class MR_VARIANT(Enum):
     MR_STRASSEN = 1
     MR_BLOCK_WIEDEMANN = 2
```
Now you can set the limits in the constructor as:
```
self.set_parameter_ranges('variant', 1, 2)
```

And with this you should be able to delete the whole function

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
```diff
@@ -0,0 +1,261 @@
...
+                temp += sum([(-1) ** (i + 1) * binomial(n, r + i) * binomial(m + i - 1, i) * binomial(K, j - i) for i in
+                             range(1, j + 1)])
+        else:
+            temp = sum(
+                [(-1) ** (i + 1) * binomial(n, r + i) * binomial(m + i - 1, i) * binomial(K + b - i - 1, b - i) for i in
+                 range(1, b + 1)])
+        return temp
+
+
+    def _dimension(self,q, n, K, r, b):
```
Memphisd, line 142: Without in-depth knowledge:
Is temp really the dimension or is it rather log(temp)? It just looks like this value might get pretty large and the general convention is passing the logarithms in those cases, unless there are good reasons not to

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
```diff
@@ -0,0 +1,261 @@
...
+        return temp
+
+
+    def _are_parameters_invalid(self, parameters: dict):
+        """
+        Specifies constraints on the parameters
+        """
+        a = parameters["a"]
+        lv = parameters["lv"]
+        b = parameters["b"]
```
Memphisd, line 159: In case we need multiple values of a dictionary like here, I think we should use

par = SimpleNamespace(**parameters)
to access the variables like par.a, par.lv etc.

(it requires "from types import SimpleNamespace")

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
Memphisd: I agree, unless there is a misunderstanding why this function was overidden.

Memphisd: Since there are many following distinctions made based on this, I am wondering if it would make sense to just create two algorithms for this -- SM using Block_Wiedemann and SM using Strassen. Or is one always superior to the other for a given set (in terms of time and memory)?

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
```diff
@@ -0,0 +1,261 @@
...
+        INPUT:
+
+        - ``parameters`` -- dictionary including the parameters
+
+        """
+        a = parameters[MR_NUMBER_OF_KERNEL_VECTORS_TO_GUESS]
+        lv = parameters[MR_NUMBER_OF_COEFFICIENTS_TO_GUESS]
+        nprime = parameters[MR_REDUCED_NUMBER_OF_COLUMNS]
+        b = parameters[MR_LINEAR_VARIABLES_DEGREE]
+        variant = parameters[MR_VARIANT]
```
Memphisd, line 228: this is inconsistent with the previous usage of "a", "lv" etc. instead of MR_NUMBER_OF_KERNEL_VECTORS_TO_GUESS...

I see the advantage of the here used way of defining a, lv, nprime etc. over the SimpleNamespaces, as it reminds the user / developer what a, lv etc. was. In that case I would be fine with either variant.

cryptographic_estimators/MREstimator/mr_algorithm.py
```diff
@@ -0,0 +1,101 @@
...
+        super(MRAlgorithm, self).__init__(problem, **kwargs)
+        w = kwargs.get("w", 2.81)
+        theta = kwargs.get("theta", 2)
+        self._name = "BaseMRAlgorithm"
+
+        if  w < 2 or 3 < w:
+            raise ValueError("w must be in the range 2 <= w <= 3")
+
+        if theta is not None and (theta > 2 or theta < 0):
+                raise ValueError("theta must be None or an integer in the range 0 <= theta <= 2")
```
Memphisd, line 42: How can theta be None, in line 35 it is set to the default value (2) or are you expecting the user to actively pass theta=None?

cryptographic_estimators/MREstimator/mr_algorithm.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        Base class for MR algorithms complexity estimator
+
+        INPUT:
+
+        - ``problem`` -- MRProblem object including all necessary parameters
+
+        """
+        super(MRAlgorithm, self).__init__(problem, **kwargs)
+        w = kwargs.get("w", 2.81)
```
Memphisd, line 34: self._w is never set to w, or am I missing it?

cryptographic_estimators/MREstimator/mr_algorithm.py
```diff
@@ -0,0 +1,101 @@
...
+        q_reduced = q
+        m_reduced = m
+        n_reduced = n - a
+        k_reduced = k - a * m - lv
+        r_reduced = r
+        return q_reduced, m_reduced, n_reduced, k_reduced, r_reduced
+
+    def cost_reduction(self, a, lv):
+        """
+        Return the cost of cost of computing the reduced instance, i.e., the obtained instance after one guess of
```
Memphisd, line 79: duplication of "cost of" in the comment

cryptographic_estimators/MREstimator/mr_algorithm.py
```diff
@@ -0,0 +1,101 @@
...
+        return max(w * log2(min(K, a * m)), log2(lv * m * n))
+
+    def hybridization_factor(self, a, lv):
+        """
+        Return the logarithm of the number of reduced instances to be solved
+
+        INPUT:
+
+        - ``a`` -- no. of vectors to guess in the kernel of the low-rank matrix
+        - ``lv``no. of entries to guess in the solution vector
```
Memphisd, line 98: double dash missing after ``lv``

cryptographic_estimators/MREstimator/mr_helper.py
```diff
@@ -0,0 +1,27 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ****************************************************************************
+
+from math import log2
+
+def _strassen_complexity_(rank, ncols):
+    """Returns the complexity of Gaussian elimination using Strassen algorithm"""
```
Memphisd, line 21: newlines missing

cryptographic_estimators/MREstimator/mr_helper.py
```diff
@@ -0,0 +1,27 @@
...
+    w = 2.81
+    return log2(7 * rank)  + (w - 1) * log2(ncols)
+
+def _bw_complexity_(row_density, ncols):
+    """Returns the complexity of block Wiedemann to find elements in the kernel of a matrix"""
```
Memphisd, line 26: same here

cryptographic_estimators/MREstimator/mr_algorithm.py
```diff
@@ -0,0 +1,101 @@
...
+        theta = kwargs.get("theta", 2)
+        self._name = "BaseMRAlgorithm"
+
+        if  w < 2 or 3 < w:
+            raise ValueError("w must be in the range 2 <= w <= 3")
+
+        if theta is not None and (theta > 2 or theta < 0):
+                raise ValueError("theta must be None or an integer in the range 0 <= theta <= 2")
+
+        self.problem.theta = theta
```
Memphisd, line 44: That could be confusing: Theta is a property of the problem object, but you are able to set it when creating an algorithm object. It could give the impression that the theta value between algorithms included in the MREstimator can vary, but in fact changing it for one algorithm, will change it for all, as all are based on the same problem object.

Memphisd: Unless you see a great advantage in usability I think I would make this property only accessible via the problem object

cryptographic_estimators/MREstimator/mr_problem.py
```diff
@@ -0,0 +1,242 @@
...
+        TESTS::
+
+            sage: from cryptographic_estimators.MREstimator.mr_problem import MRProblem
+            sage: MRP = MRProblem(q=7, m=9, n=10, k=15, r=4)
+            sage: MRP.target_rank()
+            4
+        """
+        return self.parameters[MR_TARGET_RANK]
+
+    def get_problem_parameters(self):
```
Memphisd, line 217: there is a similar function defined in the BaseProblem class called get_parameters(). would it make sense to reuse it or if it doesn't suit your purpose exactly override it?

Memphisd (changes requested):

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
Javierverbel: @FloydZ  I fixed it. The only problem with this approach is that in the printed table you see `'variant': 1` instead of the more clear one `variant = strassen`

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
Javierverbel: Notice that `* binomial(x, y)` are inside a sum. That is one has `sum([* binomial(x, y)])`. How to apply `logs` in there?

cryptographic_estimators/MREstimator/mr_algorithm.py
Javierverbel: yes, `tetha = None` means to capture the specific case (commonly) used by the community where the bit-conversion factor is set to be `2*log2(q) **2 + log2(q)`

cryptographic_estimators/MREstimator/mr_algorithm.py
Javierverbel: It was missing.

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
FloydZ: extend the `MR_VARIANT` class with:
```python
def __str__(self):
   return self.name
```
and the name should be printed

cryptographic_estimators/MREstimator/MRAlgorithms/support_minors.py
Memphisd: I mean if it is fast enough, all good. Otherwise the question is how important this level of accuracy is and if we could provide a faster variant via some approximation.

cryptographic_estimators/MREstimator/mr_algorithm.py
Memphisd: ok, hope that is documented somewhere? 😅

PR #100, opened 2023-11-02
Fix/time complexity and speed lokshtanov

Javierverbel:
### Description
- Remove unnecesary factor in the time complexity.
- Speed up the estimation
- Bring back previously removed pytests


tests/module/multivariate_quadratic_estimator/mpkc/algorithms/f5.py
```diff
@@ -97,14 +97,14 @@ def time_complexity(self, **kwargs):
             sage: from mpkc.algorithms import F5
             sage: F5_ = F5(n=10, m=15, degrees=[3]*15, q=31)
             sage: F5_.time_complexity()
-            128005423260
+            35458566000
```
Memphisd, line 100: Is it on purpose that those values are not returned as logarithm?

Memphisd (approved):

FloydZ (approved):

PR #106, opened 2023-11-20
Fix: method ncols_in_preprocessing_step method in Crossbred

Javierverbel:
### Description
- Fix method to compute the number of columns of the main matrix involved in the prepocessing step of Crosbred.


Memphisd (approved):
Changes to the code are solid. Cannot judge the change from an algorithmic view. Should be part of the changelog for the next release

PR #108, opened 2023-11-21
fix: disable variable ranges once parameters are preset

Memphisd:
### Description
- fixing algorithm parameters in the SD case let to a infinite recursion in some cases
- fixing happens by setting lower and upper bound of the parameter range to the fixed value
- the problem is related to the optimization being allowed to adapt parameter ranges to find more optimal values
- however in case of fixed values parameter ranges should not be allowed to be adapted by the optimization
- therefore this commit disables the variable parameter ranges after parameters of an algorithm have been manually fixed to some value


cryptographic_estimators/base_algorithm.py
```diff
@@ -338,7 +340,7 @@ def _fix_ranges_for_already_set_parameters(self):
         """
         parameters = self._optimal_parameters
         ranges = self._parameter_ranges
-        new_ranges = {i: ranges[i].copy() if i not in parameters else {"min": parameters[i], "max": parameters[i]}
+        new_ranges = {i: ranges[i].copy() if i not in parameters else {"min": parameters[i], "max": parameters[i] + 1}
```
FloydZ, line 343: the original fix was not complete. We need to increase the upper bound by 1, s.t. `_valid_choices` iterates at least once.

Memphisd: This change is not correct, the upper bound is generally handled as inclusive. If an SDAlgorithm overwrites the _valid_choices function it must take that into account. I see that many SD algorithms do not do this. This has to be changed instead. Otherwise algorithms of other estimators will consider to large ranges now.

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_plus.py
```diff
@@ -165,9 +165,7 @@ def _valid_choices(self):
         for p in range(new_ranges["p"]["min"], min(w // 2, new_ranges["p"]["max"]), 2):
             for l in range(new_ranges["l"]["min"], min(n - k - (w - 2 * p), new_ranges["l"]["max"])):
                 for p1 in range(max(new_ranges["p1"]["min"], (p + 1) // 2), new_ranges["p1"]["max"]):
-                    L1 = log2(binom((k+l)//2, p1))
-                    d1 = self._adjust_radius
-                    for l1 in range(max(int(L1)-d1, 0), int(L1)+d1):
+                    for l1 in range(new_ranges["l1"]["min"], min(l, new_ranges["l1"]["max"])):
```
Memphisd, line 168: why was this removed? For efficiency it is quite essential. I suggest the loop should just take additionally the ranges of l1 into account.

FloydZ (approved):

PR #110, opened 2023-12-01
Refactor/feat/bjorklund fast implementation

aangulog:
### Description
- Changes were made to how the optimal lambda is calculated, making it x200 fast.
- According to the source paper 'Solving Systems of Polynomial Equations over GF(2) by a Parity-Counting Self-Reduction' pp. 26:11, equation 21. Parenthesis for the time complexity of Bjorklund were wrong, leading to errors in its calculation.
T1 was modified so its value is not dependent on any outside parenthesis.
- description of all changes can be found in the following repository: https://github.com/aangulog/Crypto-TII-Probabilistic-algorithms-changes/blob/main/bjorklund/Bjorklund_changes.md

### Review process
- internal time complexity should give a new time complexity
- finding the optimal lambda should be faster

### Pre-approval checklist


cryptographic_estimators/MQEstimator/MQAlgorithms/bjorklund.py
```diff
@@ -186,16 +210,31 @@ def _find_optimal_tilde_o_parameters(self):
...
-            T1 = (n + (l + 2) * m * sum_of_binomial_coefficients(n, 2) +
-                  (n - l) * 2 ** (n - l))
-            s = 48 * n + 1
-            return s * sum_of_binomial_coefficients(n - l, l + 4) * (Bjorklund._internal_time_complexity_(l, l + 2, lambda_) + T1)
+        T = 0  # Time complexity
+        l = floor(lambda_ * n)  # n1, l from Björklund.
+        s = 48 * n + 1  # number of iterations
+        sumbin_n_2 = sum_of_binomial_coefficients(n, 2)
+        sumbin_B = sum_of_binomial_coefficients(n - l, l + 4)
+        T += 2 ** (n-l)  # Line 5, initialize array of size 2^{n-l}
```
Javierverbel, line 226: What Line 5 refers to? A pseudocode in a paper? Please add a reference to is in the dumentation corresponding to the method`_internal_time_complexity_`.

cryptographic_estimators/MQEstimator/MQAlgorithms/bjorklund.py
```diff
@@ -71,20 +74,41 @@ def lambda_(self):
...
+                self._current_minimum_for_early_abort = tmp_time #ver si este se puede usar para el early abort en _are_par
+                for i in params:
+                    self._optimal_parameters[i] = params[i]
+            
+            if tmp_time > time:
+                break
+            
+        self._current_minimum_for_early_abort = inf
+
         return self._get_optimal_parameter('lambda_')
```
Javierverbel, line 98: Here you have to consider the case when the complexity time is tildeO. Notice this is the case in the method `_get_optimal_parameter` in the base clase. 
This is why the test in line 208 fails.

Javierverbel (changes requested):
Fix the test and examples. 
Check the fast tests by running `make testfast`
Check all the tests by running `make testall`

cryptographic_estimators/MQEstimator/MQAlgorithms/bjorklund.py
```diff
@@ -186,16 +210,31 @@ def _find_optimal_tilde_o_parameters(self):
         """
         self._optimal_parameters['lambda_'] = 0.19677
 
-    @staticmethod
+    @staticmethod    
     def _internal_time_complexity_(n: int, m: int, lambda_: float):
         """
         Helper function. Computes the runtime of the algorithm for given n, m and lambda
```
Javierverbel, line 216: This should be the running time of the parity_count alrorithm or subroutine.

cryptographic_estimators/MQEstimator/MQAlgorithms/bjorklund.py
```diff
@@ -71,20 +74,41 @@ def lambda_(self):
             sage: E.lambda_()
             3/10
         """
+        
+        time = inf
```
Javierverbel, line 78: @aangulog I think  instead of writting the `_lambda` method it is better to rewrite the `find_optimal_parameters` for the particular case of `bjorklund`. It will solve the issue for the `TildeO` optimal params at the same time.

Javierverbel (changes requested):

aangulog (commented):
Issues have been corrected, except for the "This should be the running time of the parity_count alrorithm or subroutine." as I do not know which part of the code should be running as a subroutine. 
When the commit is done it displays the following message: this commit has not been pushed to the remote repository

Javierverbel:
@aangulog it seems you haven't update the remote branch. So the changes you've made are still in you local machine.

Javierverbel (approved):

PR #141, opened 2024-05-06
Feature/implement helper sage replacement functions

Dioprz:
### Description

**Main changes**

Python implementation of the `is_prime_power` and `is_power_of_two`functions, which will replace its homonymous Sage functions.

- `is_power_of_two`

  - Uses bitwise operations for faster computations.

  - Performance changes over Sage: 7x faster.

- `is_prime_power`

  -  Uses a precomputed list for small inputs, and uses numbers with the shape $6k\pm 1$ (prime candidates) to check divisibility on larger inputs.

  - Performance changes over Sage:

    - On inputs in the range of the precomputed primes: Same performance.

    - On larger inputs: About 1.5 times slower than Sage. This is a natural implication when comparing a Python function versus a C function from a highly optimized library (such as the one used by Sage, via the **PARI** library).

**Minor changes**

- Stylistic changes via Black.
- Remove one unused import.

### Review process

- **Feedback needed:** Do you agree to put this kind of functions in the `helper.py` file, or should I create a new file for that?
- Functions have been already intensively tested on my side.

### Pre-approval checklist

### DIFF
--- a/cryptographic_estimators/helper.py
+++ b/cryptographic_estimators/helper.py
@@ -15,15 +15,15 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 # ****************************************************************************
 
-
-from math import log2
 from enum import Enum
+from bisect import bisect_left
 
 
 class ComplexityType(Enum):
     """
     distinguish between normal optimisation and tilde O optimisation
     """
+
     ESTIMATE = 0
     TILDEO = 1
 
@@ -43,7 +43,7 @@ def concat_all_tables(tables):
 
 def concat_pretty_tables(t1: str, t2: str):
     """
-    Merge two columns into one 
+    Merge two columns into one
     INPUT:
 
     - ``t1`` -- first column
@@ -68,7 +68,7 @@ def _truncate(x: float, precision: int):
     - ``precision`` -- number of decimial digits to truncate to
 
     """
-    return float(int(x * 10 ** precision) / 10 ** precision)
+    return float(int(x * 10**precision) / 10**precision)
 
 
 def round_or_truncate(x: float, truncate: bool, precision: int):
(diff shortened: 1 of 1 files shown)


cryptographic_estimators/helper.py
```diff
@@ -82,6 +82,79 @@ def round_or_truncate(x: float, truncate: bool, precision: int):
...
+    509, 521, 523, 541, 547, 557, 563, 569, 571, 577, 587, 593, 599, 601, 607,
+    613, 617, 619, 631, 641, 643, 647, 653, 659, 661, 673, 677, 683, 691, 701,
+    709, 719, 727, 733, 739, 743, 751, 757, 761, 769, 773, 787, 797, 809, 811,
+    821, 823, 827, 829, 839, 853, 857, 859, 863, 877, 881, 883, 887, 907, 911,
+    919, 929, 937, 941, 947, 953, 967, 971, 977, 983, 991, 997
+]
+# fmt: on
+
+
+def is_prime_power(n):
```
Javierverbel, line 106: Maybe it is good idea to add several tests to these functions

Javierverbel (changes requested):

cryptographic_estimators/helper.py
Dioprz: Done!

Javierverbel (approved):

cryptographic_estimators/helper.py
```diff
@@ -60,20 +63,20 @@ def concat_pretty_tables(t1: str, t2: str):
...
 
     """
-    return float(int(x * 10 ** precision) / 10 ** precision)
+    return float(int(x * 10**precision) / 10**precision)
 
 
 def round_or_truncate(x: float, truncate: bool, precision: int):
     """
-    eiter rounds or truncates `x` if `truncate` is `true`
+    Eiter rounds or truncates `x` if `truncate` is `true`
```
FloydZ, line 79: Either

FloydZ (approved):
Nice work!

cryptographic_estimators/helper.py
Dioprz: Huh, I missed that error.  Thanks! I will fix this typo in my next PR, so we can merge this right now.

PR #143, opened 2024-05-09
Refactor/drop sage functions no series

Dioprz:
### Description

 - Change in most files that use Sage functions in the MQ Estimator, replacing them with native Python functions.
- Includes performance benchmarks on each commit:
    - ceil (Sage -> Python): 19x faster.
    - floor (Sage -> Python): 16x faster.
    - binomial (Sage -> Python): 7x faster.
    - is_prime_power (Sage -> Python): Slightly slower (up to 1.5x)
    - is_power_of_two (Sage -> Python): 7x faster.
- Unused imports were removed.
- Imports statements were modified to use absolute paths instead of relative paths. Motivation to this includes clarity, consistency and correctness. More information can be found here: [PEP-8 Imports subsection](https://peps.python.org/pep-0008/#imports).
- Stylistic changes applied via Black formatter.

### Review process
- Modifications are limited to the imports header. Other stylistic changes are the result of Black formatting.
- I use the functions implemented on https://github.com/Crypto-TII/CryptographicEstimators/pull/141 , so that PR should be merged first.

### Pre-approval checklist


Dioprz:
After some troubleshooting I find what is happening with pytest. I will give a
summary as well as the full description of the debugging process for
completeness. Possible solutions can be found at the end.

## Summary

Sage types use a different precision for rational values than standard python
float types. This combined with the kind of "laziness" evaluation used by Sage
(`a = floor(16/49 * 49)` will be `16` because it doesn't evaluate `16/49` in
advance, for example) makes its calculations more accurate in some cases, like
the one causing the pytest failure.

## Troubleshooting

#### 1. 
The execution of the next snippets clearly shows that the problem comes from
   a difference between Sage execution process, and Python execution process:

_Python_ 
```python
  >>> from cryptographic_estimators.MQEstimator.MQAlgorithms.bjorklund import Bjorklund
  >>> from cryptographic_estimators.MQEstimator.mq_problem import MQProblem
  >>> E = Bjorklund(MQProblem(n=100, m=50, q=2), bit_complexities=False)
  >>> E.memory_complexity()
  37.48546422866481
  ```
  _Sage_

  ```sage
  sage: from cryptographic_estimators.MQEstimator.MQAlgorithms.bjorklund import Bjorklund
  sage: from cryptographic_estimators.MQEstimator.mq_problem import MQProblem
  sage: E = Bjorklund(MQProblem(n=100, m=50, q=2), bit_complexities=False)
  sage: E.memory_complexity()
  36.48546485189783
  ```

#### 2. 
Digging into the internals of Bjorklund, I was able to build a minimal,
   self-contained file to reproduce the behavior:

  ```python
  from math import floor, log2, comb as binomial


  def sum_of_binomial_coefficients(n, l):
    if l < 0:
      raise ValueError("l must be a non-negative integer")
      return sum(binomial(n, j) for j in range(l + 1))


  i = 0
  def _internal_memory_complexity_(_n, _m, _lambda):
    global i
    i += 1
    print("Enter iteration", i, "with parameters", _n, _m, _lambda)
    if _n <= 1:
      return 0
      else:
      s = 48 * _n + 1
      l = floor(_lambda * _n)
      print(
        "return",
        _internal_memory_complexity_(l, l + 2, _lambda)
          + 2 ** (_n - l) * log2(s)
          + _m * sum_of_binomial_coefficients(_n, 2),
      )
      input()
      return (
        _internal_memory_complexity_(l, l + 2, _lambda)
          + 2 ** (_n - l) * log2(s)
          + _m * sum_of_binomial_coefficients(_n, 2)
      )
  print(log2(_internal_memory_complexity_(49, 49, 16 / 49)))
  ```

  To induce the inconsistency it is necessary to save the file with extension
  `.py`, and another one with extension `.sage`. The execution of both files
  with the `sage` command will show you the same results as the previous test. (Remember to press enter after each line)

#### 3. 
On the first two iterations we will start seeing differences between both
   executions. For the `.sage` file, the first two iterations of the
   `_internal_memory_complexity_` function are:

  ```console
  > sage Quick/debug/prueba.sage
  Enter iteration 1 with parameters 49 49 16/49
  Enter iteration 2 with parameters 16 18 16/49
  ```

  For the `.py` file, they are:

  ```console
  > sage Quick/debug/prueba.py
  Enter iteration 1 with parameters 49 49 0.32653061224489793
  Enter iteration 2 with parameters 15 17 0.32653061224489793
  ```

#### 4. 
From what the code is doing, is enough to understand how is working the
   definition of `l` from the first iteration to the second iteration to know
   what is happening. We conclude that the problem is the underground use of
   sage types when using a `.sage` file, as well as its associated precision.

  **Python**

  ```python
  >>> from math import floor
  >>> _lambda = 16/49
  >>> n = 49
  >>> floor(_lambda * n)
  15
  >>> _lambda * n
  15.999999999999998
  ```

  **Sage**

  ```sage
  sage: from math import floor
  sage: _lambda = 16/49
  sage: n = 49
  sage: floor(_lambda * n)
  16
  sage: _lambda * n
  16
  ```

## Solutions

1. If the propagated error is not big enough to introduce major problems, we
   can safely change the test expected value.

1. If we need more precision on the floating point operations, `Decimal`
   with the appropriate values could solve the problem. Something like this
   should work

  ```sage
  >>> from decimal import *
  >>> getcontext().prec = 50
  >>> _lambda = Decimal(16)/Decimal(49)
  >>> _lambda
  Decimal('0.32653061224489795918367346938775510204081632653061')
  >>> 49*_lambda
  Decimal('16.000000000000000000000000000000000000000000000000')
  ```

  Let me know what you think.

**Comment by Javierverbel** [comment]
We thought about this issue. We suggest the custom function
```
def custom_floor(number):
    if ceil(number)-number < 1e-10:
        return ceil(number)
    return floor(number)```
    
    So we compute `custom_floor( decimal * integer)` instead of `floor( decimal * integer) `

Dioprz:
> We thought about this issue. We suggest the custom function
> 
> ```
> def custom_floor(number):
>     if ceil(number)-number < 1e-10:
>         return ceil(number)
>     return floor(number)
> ```
> 
> ```
> So we compute `custom_floor( decimal * integer)` instead of   `floor( decimal * integer) `
> ```

While I think this is a possible solution, I have to say it feels incorrect to me because:

- With that solution, we are not addressing the root cause of the problem (float type precision), but instead patching its effects (an incorrect floor calculation).

- It raises some undesirable questions about how to handle this in the future. In the worst-case scenario, if another estimator requires this kind of highly precise calculation but with `ceil` instead of `floor`, we will need to create a `custom_ceil` function as well.

- This approach reduces code clarity for future contributors, as it might not be clear why a custom function was created instead of just using the standard `floor`.

Instead, all these issues are explicitly addressed when using `Decimal` types. We only need to make explicit casts over the numbers involved in the function, which act as a flag to indicate "here we need high precision calculations." This also provides a sane pattern to fix any future similar issues with this kind of calculation.

One more thing I discovered after playing a bit with `Decimal`: this type [doesn't work well](https://stackoverflow.com/a/48865544) in operations with "float" types. This can be seen as a feature or a bug, depending on your point of view. To me, it's beneficial because it enforces the notion that if you need high precision, it must be localized to a function, and you will need to handle explicitly where it is needed and how it is handled.

I have already tried to implement the changes in Bjorklund, and the problematic function ended up looking like this (notice the need to cast from `Decimal` to `int` in the `sum`, and `float` to `Decimal` in `log2`):

```python
from decimal import Decimal, getcontext

...

def _compute_memory_complexity(self, parameters: dict):
  lambda_ = parameters["lambda_"]

  # Default prec used by Sage
  getcontext().prec = 14

  def _internal_memory_complexity_(_n, _m, _lambda):
    if _n <= 1:
      return 0
      else:
      s = 48 * _n + 1
      l = floor(_lambda * _n)
      return (
        _internal_memory_complexity_(l, l + 2, _lambda)
          + 2 ** (_n - l) * Decimal(log2(s))
          + _m * sum_of_binomial_coefficients(int(_n), 2)
      )

    n, m, _ = self.get_reduced_parameters()
    return log2(
      _internal_memory_complexity_(Decimal(n), Decimal(m), Decimal(lambda_))
    )
```

With that said, I don't have any problem implementing your suggestion if you still think it's the best approach. I will stay tuned to know what our next steps will be!

Javierverbel (approved):

PR #149, opened 2024-05-18
Refactor/replace sage galois fields functions

Dioprz:
### Description

- Refactored `is_prime_power` to improve readibility and capabilites. Change unit tests to include more comprehensive cases.
- Implemented  custom Galois field functions used in `lokshtanov`, to drop Sage functionalities. Both functions `gf_order_to_degree` and `gf_order_to_characteristic` are based on `is_prime_power`.
-  Added unit tests and docstrings with examples to both functions.
- Removed import of the Sage constructor GF.
- Replaced Infinity Sage type by inf Python type.
- Stylistic changes via Black.

### Review process
- `is_prime_power` has major changes in code structure, but the logic is the same. Please review it to make sure it is sufficiently clear.
-  Galois functions should be very easy to follow, as all they do is call `is_prime_power` and return the appropriate value.

### Pre-approval checklist

### DIFF
--- a/cryptographic_estimators/helper.py
+++ b/cryptographic_estimators/helper.py
@@ -107,7 +107,7 @@ PRIMES = [
 # fmt: on
 
 
-def is_prime_power(n):
+def is_prime_power(n, return_pair=False):
     """
     Determines if a given number is a power of a prime number.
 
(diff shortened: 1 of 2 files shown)


cryptographic_estimators/helper.py
```diff
@@ -121,57 +121,80 @@ def is_prime_power(n):
...
+            m += 1
+        return (n == 1, m)
+
+    def check_small_primes(n, max_prime):
+        if n < max_prime:
+            index = bisect_left(PRIMES, n)
+            if index < len(PRIMES) and PRIMES[index] == n:
+                return (True, (n, 1)) if return_pair else True
+        for prime in PRIMES:
+            if prime * prime > n:
```
Javierverbel, line 160: Do we really need this condition?

Javierverbel (approved):

Javierverbel (approved):

cryptographic_estimators/helper.py
Dioprz: Sorry for the delay, I think I already answer this.

Yes, after a carefully review with professor Daniel, everything looks fine. This will only work as an early return, and don't introduce any failure in the results.

Dioprz:
@Javierverbel The history of this branch had become a bit messy, so I decided to rewrite its git history from my last commit. Now the PR must be ready for merge. In case you will add something else, remember to pull before so you get the last branch structure.

Javierverbel (approved):

PR #154, opened 2024-05-27
Refactor/hilbert series with flint

Dioprz:
### Description
- Complete reimplementation of the nmonomials class to use python-flint power series instead of Sage power series.
  - Remove of the `_ring` attribute, replacing its use by the new attribute `_gen`.
  - Remove the `series` method, to be replaced by the new methods `coefficient_of_degree` and `coefficient_up_to_degree`.
  - Rename the method `first_nonpositive_integer` as `first_nonpositive_coefficient`.
- Updated all references to the Hilbert Series class to ensure compatibility with the new implementation.
- Stylistic changes via Black formatter.

### Review process
- Major changes are in the reimplementation of HilbertSeries and the used form to call the series in Crossbred.
- Any other change should be minor or stylistic.

### Pre-approval checklist

### DIFF
--- a/cryptographic_estimators/MQEstimator/series/hilbert.py
+++ b/cryptographic_estimators/MQEstimator/series/hilbert.py
@@ -65,6 +65,36 @@ class HilbertSeries(object):
         self._series_up_to_degree = self._series / (1 - x)
 
     @property
+    def _hilbert_serie(self):
+        """
+        Return the representation of the _series attribute.
+
+        EXAMPLES::
+
+            sage: from cryptographic_estimators.MQEstimator.series.hilbert import HilbertSeries
+            sage: H = HilbertSeries(5, [2]*7)
+            sage: H._hilbert_serie
+            1 + 5*x + 8*x^2 + (-14)*x^4 + (-14)*x^5 + 8*x^7 + 5*x^8 + x^9 + O(x^14)
+
+        """
+        return self._series
+
+    @property
+    def _hilbert_serie_up_to_degree(self):
+        """
+        Return the representation of the _series_up_to_degree attribute.
+
+        EXAMPLES::
+
+            sage: from cryptographic_estimators.MQEstimator.series.hilbert import HilbertSeries
+            sage: H = HilbertSeries(5, [2]*7)
+            sage: H._hilbert_serie_up_to_degree
+            1 + 6*x + 14*x^2 + 14*x^3 + (-14)*x^5 + (-14)*x^6 + (-6)*x^7 + (-1)*x^8 + O(x^14)
+
+        """
+        return self._series_up_to_degree
+
+    @property
     def nvariables(self):
         """
         Return the no. of variables
--- a/cryptographic_estimators/MQEstimator/series/hilbert.py
+++ b/cryptographic_estimators/MQEstimator/series/hilbert.py
@@ -23,7 +23,7 @@ from math import prod
 
 class HilbertSeries(object):
     """
-    Construct an instance of Hilbert series
+    Construct an instance of Hilbert series.
 
     INPUT:
 
@@ -97,7 +97,7 @@ class HilbertSeries(object):
     @property
     def nvariables(self):
         """
-        Return the no. of variables
+        Return the no. of variables.
 
         EXAMPLES::
 
@@ -112,7 +112,7 @@ class HilbertSeries(object):
     @property
     def degrees(self):
         """
-        Return a list of degrees of the polynomials
+        Return a list of degrees of the polynomials.
 
         EXAMPLES::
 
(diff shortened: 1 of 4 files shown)


Javierverbel (changes requested):
One test in `cryptographic_estimators/MQEstimator/series/hilbert.py` is failing.

Dioprz:
> One test in `cryptographic_estimators/MQEstimator/series/hilbert.py` is failing.

Fixed. I also changed an inappropriate function implementation remaining in `hilbert.py`.

cryptographic_estimators/MQEstimator/series/hilbert.py
```diff
@@ -41,27 +40,59 @@ class HilbertSeries(object):
...
     """
 
     def __init__(self, n: int, degrees: list[int], q=None):
         self._q = q
         self._nvariables = n
         self._degrees = degrees
-        self._ring = PowerSeriesRing(
-            QQ, 'z', default_prec=2*len(self._degrees))
-        z = self._ring.gen()
+        self._prec = 2 * len(self._degrees)
```
FloydZ, line 50: why twice the degree?

cryptographic_estimators/MQEstimator/series/hilbert.py
```diff
@@ -102,90 +135,106 @@ def precision(self):
             sage: H = HilbertSeries(5, [2]*7)
             sage: H.precision
             14
+
         """
-        return self.ring.default_prec()
+        x = self._gen
+        serie_repr = x.repr()
+        return int(serie_repr[serie_repr.rfind("=") + 1 : serie_repr.rfind(")")])
```
FloydZ, line 142: this looks rather error prone. Wouldn't it be easier to store the `default_prec` from the constructor in a separate variable? And the return this variable here?

cryptographic_estimators/MQEstimator/series/hilbert.py
```diff
@@ -102,90 +135,106 @@ def precision(self):
...
+
         """
-        s = self.series()
         for d in range(self.precision):
-            if s[d] <= 0:
-                return ZZ(d)
+            if self._series_up_to_degree[d] <= 0:
+                return int(d)
         raise ValueError(
-            "unable to find a nonpositive coefficient in the series")
```
FloydZ, line 184: is it good to directly raise an exception? Im not familiar with Hilbert series, is it common that a series  does not contain a nonpositive coefficient? Or is this a rather rare thing to happen?
If its common, I think its a bad Idea to directly raise an exception. Maybe one could return 0?

FloydZ: BTW: is it a good idea to raise exception in the "hot" code at all?

cryptographic_estimators/MQEstimator/series/hilbert.py
Javierverbel: The default by PowerSeries is too small. I've noticed that `self._prec = 2 * len(self._degrees)`  should be enough and still fast for most of the cases.

cryptographic_estimators/MQEstimator/series/hilbert.py
Javierverbel: You are raising a good point. I'm afraid returning 0 might lead a misinterpretations of the results of the estimator. I think if we set the "Error" case to 0, in order to avoid misinterpretations, we have to check if the output of this method is > than 0 every time we call it, and I think this is worse than raising an error. 

When calling that function the user is expecting a nonzero degree. I think it is better make it aware of if when such a degree cannot be returned.

cryptographic_estimators/MQEstimator/series/hilbert.py
FloydZ: Maybe we should document this? In a comment or something.
Because maybe one day the precision will not be enough (maybe because of big numbers). Or is this completely impossible.

cryptographic_estimators/MQEstimator/series/hilbert.py
Javierverbel: We don't expect an MQEstimator user to always to use `try catch`.

FloydZ: ok than its fine be me

cryptographic_estimators/MQEstimator/series/hilbert.py
Dioprz: Fully agree. I don't know why I forgot the 'self._prec' attribute.

Done!

cryptographic_estimators/MQEstimator/series/hilbert.py
Dioprz: @Javierverbel What do you think? 

Does the parameters have any correlation that forces `n` to grow at the same pace that `degrees`, or we may end up having problems on big `n`'s (or `q`)?

cryptographic_estimators/MQEstimator/series/hilbert.py
FloydZ: :+1:

cryptographic_estimators/MQEstimator/series/hilbert.py
Javierverbel: I would add something like  `This precision should work for most of the case since the systems excepted to be evaluated here have a number of variables which is at most the number of equations`

Dioprz: Perfect, see my last commit.

FloydZ: perfect from my side, can be merged

FloydZ (approved):

Javierverbel (approved):

PR #155, opened 2024-05-31
Feature/implement mayo attacks

fmerino21:
### Description
- Implementation of the Kipnis-Shamir, Reconciliation, Intersection, and Claw-finding attacks, including memory and time complexity, for MAYO.
- Optimal parameter `K` removed from `mayo_algorithm.py`. Computed and used only in `direct_attack.py`.

### Review process
- Review the definition and usage of the optimal parameter `K` in `direct_attack.py`

### Pre-approval checklist


cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
```diff
@@ -15,11 +15,13 @@
...
 from ..mayo_algorithm import MAYOAlgorithm
 from ..mayo_problem import MAYOProblem
 from ...MQEstimator.mq_estimator import MQEstimator
 from ...MQEstimator.MQAlgorithms.lokshtanov import Lokshtanov
+from ...base_algorithm import optimal_parameter
+from ...helper import ComplexityType
+from ..mayo_helper import _optimize_k
 from ...base_constants import BASE_EXCLUDED_ALGORITHMS
 from math import log2
 from sage.functions.other import floor
```
Javierverbel, line 25: Can we use python functions ?

cryptographic_estimators/MAYOEstimator/mayo_estimator.py
```diff
@@ -71,63 +71,87 @@ def table(self, show_quantum_complexity=0, show_tilde_o_time=0,
...
+            |                      |              |    estimate    |
+            +----------------------+--------------+-------+--------+
+            | algorithm            | attack_type  |  time | memory |
+            +----------------------+--------------+-------+--------+
+            | DirectAttack         |   forgery    | 144.8 |   99.3 |
+            | KipnisShamir         | key-recovery | 222.1 |   17.1 |
+            | ReconciliationAttack | key-recovery | 143.2 |   48.1 |
+            | IntersectionAttack   | key-recovery | 254.8 |   33.5 |
+            | ClawFinding          |   forgery    | 142.1 |  135.1 |
+            +----------------------+--------------+-------+--------+
```
Javierverbel, line 84: Can we add optimal params to this test?

Javierverbel (changes requested):

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
```diff
@@ -0,0 +1,142 @@
...
+            sage: from cryptographic_estimators.MAYOEstimator.MAYOAlgorithms.claw_finding import ClawFinding
+            sage: E = ClawFinding(MAYOProblem(n=80, m=60, o=18, k=12, q=16))
+            sage: E.memory_complexity()
+            127.038
+
+        """
+        X = self.X()
+        Y = self.Y()
+        n, m, _, k, q = self.problem.get_parameters()
+        mem_evals = log2(log2(q)) + log2(m) + X
```
FloydZ, line 139: this `+ log(log(q))` is a translation to number of bits stored, even if the complexity_type set to 0

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/intersection_attack.py
```diff
@@ -0,0 +1,104 @@
...
+
+    """
+
+    def __init__(self, problem: MAYOProblem, **kwargs):
+        super().__init__(problem, **kwargs)
+
+        self._name = "IntersectionAttack"
+        self._attack_type = "key-recovery"
+        n, m, o, _, q = self.problem.get_parameters()
+        self._boolean_solve = BooleanSolveFXL(MQProblem(n=n, m=3*m-2, q=q), bit_complexities=False)
```
FloydZ, line 52: the bit_complexity of the sub-MQProblem should be the same as the given `bit_complexity`.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/kipnis_shamir.py
```diff
@@ -0,0 +1,98 @@
...
+            sage: from cryptographic_estimators.MAYOEstimator.mayo_problem import MAYOProblem
+            sage: from cryptographic_estimators.MAYOEstimator.MAYOAlgorithms.kipnis_shamir import KipnisShamir
+            sage: E = KipnisShamir(MAYOProblem(n=24, m=20, o=8, k=9, q=16))
+            sage: E.time_complexity()
+            50.007820003461546
+
+        """
+        n, _, o, _, q = self.problem.get_parameters()
+        time = (n - 2 * o) * log2(q)
+        return time + log2(n) * 2.8
```
FloydZ, line 77: @Javierverbel I think we already spoke about this. You decided to hardcode this constant?

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/reconciliation_attack.py
```diff
@@ -0,0 +1,100 @@
...
+
+        INPUT:
+
+        - ``parameters`` -- dictionary including the parameters
+
+        TESTS::
+
+            sage: from cryptographic_estimators.MAYOEstimator.mayo_problem import MAYOProblem
+            sage: from cryptographic_estimators.MAYOEstimator.MAYOAlgorithms.reconciliation_attack import ReconciliationAttack
+            sage: E = ReconciliationAttack(MAYOProblem(n=32, m=30, o=12, k=9, q=16))
```
FloydZ, line 65: why is here a different parameter set, then in the previous algorithms?

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/reconciliation_attack.py
```diff
@@ -0,0 +1,100 @@
...
+
+    """
+
+    def __init__(self, problem: MAYOProblem, **kwargs):
+        super().__init__(problem, **kwargs)
+
+        self._name = "ReconciliationAttack"
+        self._attack_type = "key-recovery"
+        n, m, o, _, q = self.problem.get_parameters()
+        self._boolean_solve = BooleanSolveFXL(MQProblem(n=n-o, m=m, q=q), bit_complexities=False)
```
FloydZ, line 51: is BooleanSolveFX always the best?  or why is it hardcoded here. Otherwise I suggest to find the optimal solver via `MQEstimator`

FloydZ (changes requested):

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
```diff
@@ -78,6 +80,29 @@ def __init__(self, problem: MAYOProblem, **kwargs):
         
         self._fastest_algorithm = None
 
+    @optimal_parameter
+    def K(self):
+        """
+        Return the optimal parameter `K` from Furue, Nakamura, and Takagi strategy
```
Memphisd, line 86: This is not clear to me. The constructor says it is a direct attack via the MQ Estimator. Why is there an additional optimal parameter K? If this is a different strategy to obtain a MQ instance from MAYO, shouldn't this be an extra attack? Also there should be a citation towards Furue, Nakamura, and Takagi

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
```diff
@@ -78,6 +80,29 @@ def __init__(self, problem: MAYOProblem, **kwargs):
...
+            sage: from cryptographic_estimators.MAYOEstimator.mayo_problem import MAYOProblem
+            sage: E = DirectAttack(MAYOProblem(n=80, m=60, o=18, k=12, q=16))
+            sage: E.K()
+            15
+                
+        """
+        if self._optimal_parameters.get("K") is None:
+            n, m, _, k, q = self.problem.get_parameters()
+            if self.complexity_type == ComplexityType.ESTIMATE.value:
+                return _optimize_k(n=k*n, m=m, k=k, q=q)
```
Memphisd, line 100: this doesn't follow the general design by implementing the optimization routine outside the framework.

Memphisd: if there is a strong argument to do it, please specify, otherwise we should adhere to the usual framework I think

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/intersection_attack.py
```diff
@@ -0,0 +1,104 @@
...
+    - ``complexity_type`` -- complexity type to consider (0: estimate, default: 0)
+    - ``bit_complexities`` -- determines if complexity is given in bit operations or basic operations (default 1: in bit)
+
+    """
+
+    def __init__(self, problem: MAYOProblem, **kwargs):
+        super().__init__(problem, **kwargs)
+
+        self._name = "IntersectionAttack"
+        self._attack_type = "key-recovery"
```
Memphisd, line 50: this string should be a constant in the base constant file and then imported (as it has to be used in this exact form everywhere)

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/kipnis_shamir.py
Memphisd: not sure I was part of that discussion, but I also think it should be a parameter like usually (in case it is the matrix multiplication constant), no?

Memphisd (changes requested):

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
```diff
@@ -0,0 +1,142 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ****************************************************************************
+
+
+from ..mayo_algorithm import MAYOAlgorithm
+from ..mayo_problem import MAYOProblem
+from ...MQEstimator.mq_estimator import MQEstimator
+from ...MQEstimator.MQAlgorithms.lokshtanov import Lokshtanov
+from ...base_constants import BASE_EXCLUDED_ALGORITHMS
+from ...base_algorithm import optimal_parameter
```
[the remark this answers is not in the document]
Javierverbel: During the transition to python we were changing from relative to absolute. 
We will move the MQEstimator to relative imports.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
```diff
@@ -0,0 +1,142 @@
...
+
+            sage: from cryptographic_estimators.MAYOEstimator.mayo_problem import MAYOProblem
+            sage: from cryptographic_estimators.MAYOEstimator.MAYOAlgorithms.claw_finding import ClawFinding
+            sage: E = ClawFinding(MAYOProblem(n=80, m=60, o=18, k=12, q=16))
+            sage: E.X()
+            122.962
+
+        """
+        _, m, _, _, q = self.problem.get_parameters()
+        X = (1 / 2) * (17 - log2(36 * m) + m * log2(q))
```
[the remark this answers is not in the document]
Javierverbel: I agree on adding a comment explaining where these numbers comes from. 
AFAIK, there is not public pdfs explaining those. The MAYO specs states what is in the image.
When we do the math we find that the `log` of the optimal `X` is what is written in line 69.
For know we can add a comment, than latter when we implement the WiKi we add the explanation of these value there. How does it sound? @FloydZ @fmerino21 

image

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
```diff
@@ -0,0 +1,142 @@
...
+from math import log2, e
+
+
+class ClawFinding(MAYOAlgorithm):
+    """
+    Construct an instance of ClawFinding estimator
+
+    Claw finding attack is a general attack which works against any signature which 
+    follows the hash-and-sign paradigm. 
+
```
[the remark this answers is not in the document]
Javierverbel: So we wait until the Wiki in online?
Or we move forward for now? @Memphisd

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/intersection_attack.py
Javierverbel: I agree.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
Javierverbel: You're right, it should be
`mem_evals =  log2(m) + X`  to indicate the log of the number of Fq to be stored for the evaluations of the pk.
`mem_hash = log(256) Y - log(q)` to indicate log of the bit_memory needed to store the hashed interpreted as the number of field elements to store.

Also the hash length (stated as 256) depends on the target security level.
It should be 256, 384 and 512 for categories I, III, and V, respectively.
It can simply use 256 to reduce the complexity of the code? It will underestimate the bit memory by one bit (in the case (`mem_hash` = min(mem_hash, mem_evals)`) 
Or we  include a `security_parameter` for every scheme parameter set? 
In the last case we would use `mem_hash = log(2 * security_parameter) Y - log(q)` .

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/intersection_attack.py
Javierverbel: I disagree. If  it is the same as the given `bit_complexity`, then when  `bit_complexity =1` 
Then the sub-MQAlgorithmEstimator adds the `bit_complexity` factor and then external External estimator (in this case MAYOEstimator) adds the `bit_complexity` factor one more time leading to an overestimation.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/kipnis_shamir.py
Javierverbel: 1 . I decided to hardcore this constant.
2. It corresponds to the matrix multiplication constant.

The motivations to hardcode it are:
1. It is explicitly stated like this in MAYOspces.
2. This constant is also used in different parts of the estimator to estimate the complexity of a different thing (estimate the complexity of an MQ instance), where it might make more sense to set it to a different value. 

Currently, we leave the input `w` to be used in the parts of the MAYOEstimator where we have to solve an instance of the MQ problem, and we hard code It to `2.8` in kipnis-shamir where we have to find the eigenvectors of relative small matrix.

It my opinion for kipnis-shamir would me more sense to hardcode a `3` instead of `2.8` since the matrices are small (< 1000 rows and columns), but I decided to use the `2.8` since it explicitly stated like that in the MAYO specs.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/reconciliation_attack.py
Javierverbel: All of them have different parameter sets. Do you consider important to state all the examples with the same parameter set?

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/reconciliation_attack.py
Javierverbel: Good comment.

1. It is use like this in the MAYO specs.
2. Using `MQEstimator` to optimize `K` would be very slow.
3. For the parameters proposed by MAYO BooleanSolveFX is the best.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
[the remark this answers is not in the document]
Javierverbel: I agree with the idea of the wiki

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/reconciliation_attack.py
Javierverbel: For now we have decided to optimize `K` using BooleanSolveXL and estimate the complexity with the `MQEstimator` using such an optimal `K`.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
Javierverbel: I agree on adding a citation to the paper of Furue, Nakamura, and Takagi.

Yes, it is a different strategy that haven't yet been implemented In the MQEstimator.

Javierverbel: It shouldn't be an extra attack since it is how the direct_attack is defined in the MAYO specs.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
Javierverbel: Agree, it doesn't follow the general optimization routine. 
The strong argument to do it is efficiency. 
In the general framework we would need to use the `_compute_time_complexity` as the function to optimize, which in this case is the MQEstimator.fastest_algorithm function, and it is super slow. 
In addition, the MAYO estimator uses BooleanSolveXL as the optimization function so we do so. Still we use the MQEstimator.fastest_algorithm  to get time complexity estimates using the optimal `K` computed with BooleanSolveXL.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
```diff
@@ -50,10 +52,10 @@ def __init__(self, problem: MAYOProblem, **kwargs):
```
Javierverbel, line 52: To fix in the doc:
`DummyProblem` --> `MQProblem`

Javierverbel (changes requested):

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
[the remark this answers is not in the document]
Memphisd: if there is a good reason to move to absolute imports we can also re-discuss it. I remember that if you want to execute individual doctests from the IDE absolute imports make it easier... but I am not sure anymore. Lets put it on the agenda for the next meeting.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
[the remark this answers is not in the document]
Memphisd: nono, we can move forward, but we should also keep the wiki overview in mind, I'll add it to the agenda of the next meeting as well.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/kipnis_shamir.py
Memphisd: I am not convinced that hardcoding is the correct way. You are even saying that using 2.8 might be very conservative. So if you want to optimize for more aggressive MAYO parameters you would need the possibility to set the value. I would rather parameterize it with standard value 2.8 or 3. Even if this means there is a MAYO_w and a MQ_w input.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
Memphisd: Ok, good for me. I think it's not even outside the framework. It seems the K parameter can be optimized independently. In those cases we don't need to use the `get_optimal_parameters` (see eg. the `r` parameter in the `SDAlgorithm` class).

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/reconciliation_attack.py
FloydZ: > For the parameters proposed by MAYO BooleanSolveFX is the best.
then its fine

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/kipnis_shamir.py
Javierverbel: Ok sounds reasonable.
@fmerino21 can we add as parameters one `w` for the MQ instances of MAYO and one `w` for the Kipnis-Shamir attack. Names can be `w_mq` and `w_kipnis_shamir`, where the  `w_kipnis_shamir` is set by default to 2.8?

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
Javierverbel: At the beginning we implemented it as `r` in `SDAlgorithm` but then we realized this was not efficient since `K` is only used for the `direct_attack` (opposite as `r`). Hence, no need  to optimized when you want to estimate other attack, for instance, `kipnis_shamir`.

@fmerino21 could you confirm if this is correct?

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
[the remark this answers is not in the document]
Dioprz: Yes. Absolute imports are:

- The default standard as can be found in the [PEP 8 – Style Guide for Python Code](https://peps.python.org/pep-0008/#imports)
- Much more clearer than a relative import: If I have a `helper.py` in each subdirectory, it's hard to know where this comes from
    ```python
     from ...helper import is_prime_power
  ```
  But is easier if we have, instead:
  ```python
  from cryptographic_estimators.helper import is_prime_power
  ```
- While the previous mentioned Style Guide also mentions that relative imports are an acceptable alternative *..."especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose"*, our imports are really not such deep to make them unnecesarily verbose and I don't think they will be in the future; as the nesting of our modules are kind of limited by the structure we are following (`cryptographic_estimators.[estimator].[algorithms].[algorithm]`, in the worst case).

I have made the changes just in the files I have acces to in the Sage migration process, but I plan to make this convention explicit in the documentation and then make the changes in the rest of the files (if we agree about it, of course).

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
fmerino21: Exactly as Javier said, it is not efficient to compute `K` for all algorithms given that we are only using it in the `direct_attack`. Besides that, by implementing the `K` parameter that way, when we use the method `.table(show_all_parameters=1)` in an instance of the `MAYOEstimator`, the `K` parameter is shown in all the algorithms, which makes no sense.

cryptographic_estimators/UOVEstimator/UOVAlgorithms/kipnis_shamir.py
```diff
@@ -20,6 +20,7 @@
 from ..uov_problem import UOVProblem
 from math import log2
 from sage.arith.misc import is_power_of_two
```
Javierverbel, line 22: Please use the python function `is_power_of_two` which is in `helper.py`

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
```diff
@@ -34,8 +35,8 @@ class DirectAttack(MAYOAlgorithm):
 
     INPUT:
 
-    - ``problem`` -- DummyProblem object including all necessary parameters
-    - ``w`` -- linear algebra constant (default: Obtained from MAYOAlgorithm)
+    - ``problem`` -- MAYOProblem object including all necessary parameters
+    - ``w_mq`` -- linear algebra constant (default: Obtained from MAYOAlgorithm)
```
Javierverbel, line 39: `obtained from MAYOAlgorithm` lower case the `o`

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
```diff
@@ -0,0 +1,151 @@
...
+        # hash_bit_size can be adjusted to target a specific security level, as stated in MAYO specificactions
+        # (security level 1: 256, security level 3: 384, security level 5: 512)
+        self._hash_bit_size = kwargs.get("hash_bit_size", 512)
+
+    @optimal_parameter
+    def X(self):
+        """
+        Return logarithm of the optimal `X`, i.e. no. of inputs (preimages)
+        Optimal value for `X` is obtained from optimizing the bit-cost expression of the attack 
+        reported in MAYO specification 36 * m * X + Y * 2 ** 17 using X * Y = q ** m
```
Javierverbel, line 64: Can we add a reference here? something like `(see Section X.X in [MAYOSpecs]`?

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/reconciliation_attack.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Construct an instance of ReconciliationAttack estimator
+
+    Reconciliation attack attempts to find vectors in the oils space O by using the fact
+    that P(o) = 0 for all o in O. 
+
+    INPUT:
+
+    - ``problem`` -- MAYOProblem object including all necessary parameters
+    - ``w`` -- linear algebra constant (default: Obtained from MAYOAlgorithm)
```
Javierverbel, line 36: `obtained from MAYOAlgorithm`

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    Construct an instance of ClawFinding estimator
+
+    Claw finding attack is a general attack which works against any signature which 
+    follows the hash-and-sign paradigm. 
+
+    INPUT:
+
+    - ``problem`` -- MAYOProblem object including all necessary parameters
+    - ``w`` -- linear algebra constant (default: Obtained from MAYOAlgorithm)
```
Javierverbel, line 38: The default should be the same as in the MAYOEstimator which currently is stating that the default value of `w` is `2`, this value should be 2.81 by default in the MAYOEstimator

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/kipnis_shamir.py
```diff
@@ -0,0 +1,98 @@
...
+    Construct an instance of KipnisShamir estimator
+
+    First attack on the Oil and Vinegar problem, proposed by Kipnis and Shamir.
+    The attack attempts to find vectors in the oil space O, by exploiting the fact
+    that these vectors are more likely to be eigenvectors of some publicy-known matrices.
+
+    INPUT:
+
+    - ``problem`` -- MAYOProblem object including all necessary parameters
+    - ``w_ks`` -- linear algebra constant (default: 2.8)
```
Javierverbel, line 36: `linear algebra constant (only for kipnis-shamir algorithm)`
This parameter should also appear in the documentation of the MAYOEstimator.

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/direct_attack.py
Javierverbel, line 39: Since it is called `w` all the other algorithms I suggest to leave it as `w` only

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/kipnis_shamir.py
Javierverbel: In order to explain the choice of 2.8 as the default value of `w_ks`.
We can add a note right before `INPUT:`

Something like:
```
NOTE::
The linear algebra constant `w_ks` as set by default to 2.8 since this is the suggested choice in Section X.X of [MAYOspecs]_ 
```

cryptographic_estimators/MAYOEstimator/mayo_estimator.py
Javierverbel, line 1: `w -- linear algebra constant (default: 2)` ---> `w -- linear algebra constant (default: 2.81)`

Javierverbel (changes requested):

cryptographic_estimators/UOVEstimator/UOVAlgorithms/kipnis_shamir.py
fmerino21: I deleted that import because it wasn't being used

Javierverbel (dismissed):

Javierverbel:
It looks to me.

cryptographic_estimators/MAYOEstimator/mayo_estimator.py
```diff
@@ -66,68 +67,94 @@ def table(self, show_quantum_complexity=0, show_tilde_o_time=0,
...
-            | DirectAttack |   forgery   | 21.4 |   56.9 | {'K': 10}  |
-            +--------------+-------------+------+--------+------------+
-
-            sage: E = MAYOEstimator(n=64, m=60, o=10, k=21, q=16)
+            +----------------------+--------------+----------------------------------------------------+
+            |                      |              |                      estimate                      |
+            +----------------------+--------------+-------+--------+-----------------------------------+
+            | algorithm            | attack_type  |  time | memory |             parameters            |
+            +----------------------+--------------+-------+--------+-----------------------------------+
+            | DirectAttack         |   forgery    |  21.4 |   56.9 |             {'K': 10}             |
```
FloydZ, line 138: This looks wrong, the numbers are far to good, and memory > time

FloydZ (changes requested):

cryptographic_estimators/MAYOEstimator/mayo_estimator.py
Javierverbel: Nice catch!

@fmerino21 I checked the problem seems to be that. 
In the for the memory complexity we shouldn't add the factor `q^K`, where `K` is the number of variables to specialize.  Here we should assume that to solve each subsystem with `n-K` variables  the same  memory space  is used.

For instance, see how is handled in `hybrid_f5.py`

FloydZ (approved):

Javierverbel (approved):

cryptographic_estimators/MAYOEstimator/MAYOAlgorithms/claw_finding.py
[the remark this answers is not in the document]
Memphisd: alright lets quickly touch it in the next meeting and in case there are no good reasons to do otherwise we can move to absolute imports (everywhere)

Memphisd (approved):
Looks good to me. We should remember to touch the points of: 1) Imports and 2) Documentation of basic operations, in one of the next meetings.

PR #156, opened 2024-06-10
Refactor/revamp tests dir structure

Dioprz:
### Description

First step in the refactor of the tests structure.

Changes include:
- Now the tests directory is a 1-1 mirror of the cryptographic_estimators directory structure. Each estimator will have a main test handler file (named `test_<estimator_name>.py`), and could have a directory with legacy implementations that we want to reuse as tests.
- Split a file that tested multiple estimators simultaneously on their respective estimators: `tests/test_estimators.py`
- Change the undescriptive directory name `module` to `helpers` and update all import references to its files.
- Clean the directory structure of the MQEstimator (legacy implementation).
- Improve import declarations and general code style via Black.
- Update the testing paths in the `make docker-pytest` command.
- **Add a new dependency for pytest**. This dependency allows pytest to use **multicore** test execution, improving *drastically* the time that takes (and will take) run the command.
    - This dependency is used in the `make docker-pytest` with the option `-n auto`, meaning that the command will take all the system cores by default, but not all tests are parallelizable, so we don't take 100% of the cores all the time.

### Review process
I have already run all the tests on my side to ensure that we don't have any missing references, but a general overview might be nice.

### Pre-approval checklist


tests/MQEstimator/test_mq.py
```diff
@@ -0,0 +1,168 @@
...
+test_sets = [
+    # n   m   q
+    [50, 50, 2],
+    [70, 70, 4],
+    [50, 70, 8],
+    [120, 40, 8],
+]
+
+algos = [
+    Bjorklund,
```
FloydZ, line 48: This is a little bit strange. Whenever you want to add a new MQAlgorithm, you need to this at 4 different places, only within this file. Is there a way to enumerate over the imports of an import statement, so this array is build up dynamically?

tests/MQEstimator/test_mq.py
```diff
@@ -0,0 +1,168 @@
...
+        F5,
+        KPG,
+        CGMTA,
+        MHT,
+    ]
+
+    mq_estimator = MQEstimator(100, 50, 2, excluded_algorithms=excluded_algorithms)
+    result = mq_estimator.estimate()
+    expected_result = {
+        "Bjorklund": {
```
FloydZ, line 95: Isn't it better to store those results in a file `results.json`? This would make the update process much easier, if we want to add more tests for example.

tests/MQEstimator/test_mq.py
```diff
@@ -0,0 +1,168 @@
...
+        Lokshtanov,
+        Crossbred,
+        F5,
+        KPG,
+        CGMTA,
+        MHT,
+    ]
+
+    with pytest.raises(TypeError, match="q must be equal to 2"):
+        Bjorklund(MQProblem(100, 50, 4))
```
FloydZ, line 131: the exception, should be already risen here. So why the next two lines?

tests/SDEstimator/test_sde.py
```diff
@@ -0,0 +1,176 @@
...
+        BJMMdw,
+        BothMay,
+        MayOzerov,
+        Stern,
+    ]
+    sd_estimator = SDEstimator(100, 50, 2, excluded_algorithms=excluded_algorithms)
+    result = sd_estimator.estimate()
+    expected_result = {
+        "Prange": {
+            "additional_information": {
```
FloydZ, line 72: the same as in the mqestimator

Dioprz:
@FloydZ  Huh, sorry 😞 , I don't mention in the PR message the intention of this structure refactor.

Every change you see in the PR is either:

1. One of the explicitely mentioned in the PR description.
2. An stylistic change with Black.
3. An import change.

I left the rest of the code as it was, because the next step is going estimator by estimator cleaning the code in detail and adding the new testing logic.

This is what I was referring to by saying "the first step", but sorry, I was not explicit, my bad.

Most of the comments you mention will be addressed in one of the next PR's (or that was my intention. If you think I should make all the changes in just one PR... It will be really hard to review but I can do it)

FloydZ:
> @FloydZ Huh, sorry 😞 , I don't mention in the PR message the intention of this structure refactor.
> 
> Every change you see in the PR is either:
> 
>     1. One of the explicitely mentioned in the PR description.
> 
>     2. An stylistic change with Black.
> 
>     3. An import change.
> 
> 
> I left the rest of the code as it was, because the next step is going estimator by estimator cleaning the code in detail and adding the new testing logic.
> 
> This is what I was referring to by saying "the first step", but sorry, I was not explicit, my bad.
> 
> Most of the comments you mention will be addressed in one of the next PR's (or that was my intention. If you think I should make all the changes in just one PR... It will be really hard to review but I can do it)

ah ok, then ignore my remarks. But I'm still intrigued to find out how to enumerate the imports of an import statement.

Dioprz:
Yes, that is possible using `import_module` from `importlib`.

Something like this has  the same effect as `from A import B as C`.

```
import importlib

module_name = "A"  # Module to import from
attribute_name = "B"  # Attribute (class, function, etc.) to import
alias = "C"  # The new name to use

module = importlib.import_module(module_name)
globals()[alias] = getattr(module, attribute_name)  
```

Javierverbel (dismissed):

Javierverbel (changes requested):
I get the following error while executing `make docker-pytest`

`docker: Error response from daemon: Conflict. The container name "/pytest-estimators" is already in use by container "c502569e9caffd314a8d5f3c38971a34874871a6cfc7d217c683b6d8b6e9aea6". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
make: *** [docker-pytest] Error 125
`

Dioprz:
Yes, that error comes when your previous execution of the command have been interrupted, or it failed. You can fix it with:

```
make stop-container-and-remove container_name="pytest-estimators"
make docker-pytest
```

Is very easy to end up having this error, so let me send a commit that fix it.

Dioprz:
Done. Please try to pull the new commit and run the command again.

Changes in the code will ensure two things:

1. After this command is executed, you will never end up with an "pytest-estimators" container active.
2. Even if you have created that container with a manual execution (maybe in a debugging process or something like that), this command will remove your previous container, and then create a new (updated) one.

Javierverbel:
> Done. Please try to pull the new commit and run the command again.
> 
> Changes in the code will ensure two things:
> 
> 1. After this command is executed, you will never end up with an "pytest-estimators" container active.
> 2. Even if you have created that container with a manual execution (maybe in a debugging process or something like that), this command will remove your previous container, and then create a new (updated) one.

Dioprz:
Thanks for pointing that out. For some reason I think that Make commands were not interrumpted if any of its lines failed (something like the behaviour with pipelines `|`)... But no.

I was testing a common fix pattern for this situations, using the OR operator:

```
docker-pytest:
	@make stop-container-and-remove container_name="pytest-estimators" || {echo "No container found for removal."}
	@docker run --name pytest-estimators -d -it ${image_name} sh \
		&& docker exec pytest-estimators sh -c "sage --python3 -m pytest -n auto -vv \
		--cov-report xml:coverage.xml --cov=${PACKAGE} \
		&& ${SAGE} tests/SDFqEstimator/test_sdfq.sage \
		&& ${SAGE} tests/LEEstimator/test_le_beullens.sage \
		&& ${SAGE} tests/LEEstimator/test_le_bbps.sage \
		&& ${SAGE} tests/PEEstimator/test_pe.sage \
		&& ${SAGE} tests/PKEstimator/test_pk.sage" || \
		{echo "Tests have failed."}
	@make stop-container-and-remove container_name="pytest-estimators"

```

However, I also was planning to propose some changes in the Makefile commands to make them clearer.

About your solution, I liked it but with one concern. This pattern can make handling errors in multiline commands more complex. because then each line could potentially imply a if/else validation.

Altought I don't think my solution is better (it is less expressive for something unfamiliar with bash scripting, for example), is less verbose.

So... Yeah, any solution will be an improve, so let me know which do you prefer and I will send you a commit in a couple minutes to solve this.

Dioprz:
> Even when the image is open and I run `make docker-pytest` I get the following error
> 
> ```
> pytest-estimators
> pytest-estimators
> dcd877c2e763871d8e9c92419fb7193605beea0b3d44af07e60ad9555f9b00c3
> ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
> __main__.py: error: unrecognized arguments: -n
>   inifile: None
>   rootdir: /home/cryptographic_estimators
> 
> make: *** [docker-pytest] Error 4
> ```

I have not been able to reproduce this error. Any other clues as to how I can induce it?

Dioprz:
I send a commit with my proposal so that in case you like it, we can make the merge. Otherwise, I can implement your suggestion at  your signal.

The other commit includes path import updates on two files I previously don't see (and I think tests was not calling).

Javierverbel:
> I send a commit with my proposal so that in case you like it, we can make the merge. Otherwise, I can implement your suggestion at your signal.
> 
> The other commit includes path import updates on two files I previously don't see (and I think tests was not calling).

You solution is fine for me: 

Another issue is: When I execute `make docker-pytest` 
I get 

```4ccf0b82bfa47c9e19074564779694668bf345055393ece8468266c2ec81c8b4
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: -n
  inifile: None
  rootdir: /home/cryptographic_estimators

Some test have failed. Please see previous lines. Cleaning container...
```

Some minor suggestions 

 `"Container removed. Continuing..."` --->  `Removing previous container...`. 
`No container had to be removed. Continuing..` ---> `Building docker image` or `Creating container...`

Javierverbel (dismissed):
After minor comments it is ok to merge from my side.

Dioprz:
Apologies for the delay.

Your suggestions are applied in a way that makes sense with the process being done in the execution flow.

For the record, an as we discussed,  the error you encountered can be resolved by rebuilding the Docker image. Therefore, to test the commit, please run the following commands:

```fish
make docker-build
make docker-pytest
```

Javierverbel (approved):
It looks good to me.

FloydZ (approved):

PR #161, opened 2024-07-05
Refactor/new regression tests architecture

Dioprz:
### Description

This PR includes the new implementation of the **regression tests** architecture, based on the next principles:

-  Preservation of the reference Sage code we have to test our estimators implementations.
-  Ability to provide new Sage code only if it will be used for testing purposes.
-  Precompute reference values to drastically improve our testing times.
-  Another step in the process of remove the dependency on Sage into the code intended for end users.

**Implemented changes include**:

**Logic and testing infrastructure**
- Creation of a `REFERENCES_INPUTS` dictionary where to define which inputs will be provided to the regression tests.
- Implementation of the logic to parse that dictionary, generate expected tests outputs on those inputs, and then serialize both inputs and outputs.
- Implementation of a custom-made python import function capable to import from Sage modules.
- Add support for `.sage` and `.py` file generators.
- Creation of a new docker command `docker-generate-tests-references` to create (or overwrite) the serialized references `.yaml` file.
- Implementation of the pytest logic used to efficiently load the serialized data, based on [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) and declared in the `conftest.py` file.

**Test migration**
- First estimator migration to the new testing process applied to `SDFqEstimator`, demonstrating an approximate 17x speed improvement (~53 secs -> ~3 secs).

**Documentation**
- Add a comprehensive documentation via docstrings in all the new code.
- Creation of a new testing documentation section explaining how it works now, as well as a guide on how to implement a regression test in the library.

### Review process
- First, read the new documentation section to get a solid grasp on the internals introduced by this PR: [`/docs/tests.md`](https://github.com/Crypto-TII/CryptographicEstimators/blob/refactor/new_regression_tests_architecture/docs/tests.md#tests). If something is not clear enough, please let me know.

- Next, follow these steps to test the new infrastructure:
   ```shell
  make docker-build
  make docker-generate-references #the PR already comes with a precomputed file at `tests/validations/reference_values.yaml`, but you can delete it to be sure everything is fine

  make docker-run
  pytest tests/validations/ #inside the docker container
   ```
  This commands sequence will:
   - Run the complete generators logic (first and second testing phase) on the three SDFqEstimator declared generators.
   - Run the tests from the serialized data on the corresponding SDFqEstimator tests functions.

### Pre-approval checklist


Javierverbel (approved):
After few minors comments it looks fine to me.

Javierverbel:
@Dioprz let's wait for @Memphisd or/and @FloydZ approval to merge this one.

tests/references/generate_references.py
```diff
@@ -0,0 +1,194 @@
...
+
+    for gen_path, outputs in gen_paths_and_inputs_with_outputs:
+        gen_info = cast(dict, get_generator_info(reference_data, gen_path))
+        del gen_info["inputs"]
+        gen_info["inputs_with_expected_outputs"] = outputs
+
+    reference_data = flatten_nested_dict(reference_data)
+    reference_data = rename_gen_to_test(reference_data)
+
+    # We only write the reference data into a file if its serialization
```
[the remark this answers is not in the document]
Dioprz: What do you think now?

docs/tests.md
```diff
@@ -0,0 +1,345 @@
...
+
+     - Takes a single parameter, `test_data`, automatically provided by _pytest_
+       and containing the input/output pairs from the `reference_values.yaml`
+       file.
+     - The boilerplate code is expected to have the following structure:
+
+     ```python
+     def test_<my_estimator>_<my_function>(test_data):
+         """
+         Test for the ... problem with an error tolerance of epsilon.
```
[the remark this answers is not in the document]
Dioprz: Docstrings should avoid hardcoded values. This prevents contributors from having to manually update the information in both the docstring and the code. Since this value can vary between tests, referencing the `epsilon` variable defined on the next line is a more maintainable approach.

Dioprz:
Thanks for the hints @Javierverbel!

I've incorporated your changes and added another based on Prof. Daniel's suggestion: I've moved the `REFERENCE_INPUTS` dictionary to a separate file to prevent accidental modifications to the `generate_references.py` logic. I think this approach is cleaner, and I've also updated the documentation accordingly.

Javierverbel (dismissed):

tests/references/SDFqEstimator/gen_sdfq.sage
```diff
@@ -0,0 +1,101 @@
...
+        A list of tuples, each containing the input parameters and the corresponding expected complexity.
+    """
+
+    def gen_single_case(input):
+        n, k, w, q = input
+        expected_complexity, _, _ = peters_isd(n, k, q, w)
+        return input, expected_complexity
+
+    inputs_with_expected_outputs = list(map(gen_single_case, inputs))
+    return inputs_with_expected_outputs
```
Memphisd, line 63: Could this maybe be handled more elegantly by providing a generic function that produces the test cases once ``gen_single_case`` and the algorithm name is provided?

tests/references/generate_references.py
```diff
@@ -0,0 +1,175 @@
...
+        import_and_execute_generator, gen_paths_and_inputs
+    )
+
+    for gen_path, outputs in gen_paths_and_inputs_with_outputs:
+        gen_info = cast(dict, get_generator_info(reference_data, gen_path))
+        del gen_info["inputs"]
+        gen_info["inputs_with_expected_outputs"] = outputs
+
+    reference_data = flatten_nested_dict(reference_data)
+    reference_data = rename_gen_to_test(reference_data)
```
Memphisd, line 162: For what do we need the "gen_[...]" and test_[...]" distinction in the dictionary? Why not just remove that layer, the geneneration then searches for the "inputs" and generates the "outputs", the test step later searches for both and compares against our estimator.

Memphisd (changes requested):
The overall framework looks good. Not sure if we can maybe lighten it a bit to avoid some of the code duplications (as indicated by my comment).

Generally, there should be make commands not relying on docker for those having sage installed locally, i.e. make generate-tests-references etc.

tests/references/generate_references.py
Dioprz: There are a couple of reasons to that:

1. What we are describing in the dictionary is the full path, from the `cryptographic_estimators/tests/` directory, to the file containing the generator. We need that path to check if the file has a `.sage` or a `.py` extension, and then define which import method we need to use.

2. Generators have a completely different logic than tests. So we need to be able to describe, without ambiguity, if we will use the `sdfq_lee_brickell` **generator** function, or the `sdfq_lee_brickell` **test** function. That also applies to the files containing the functions, otherwise we will end having two files with the same name, one in validations and one in references. Easiest way I found to create that distinction was with the `gen` and `test` prefix.

3.  With that said, it's perfectly possible to remove the `gen` prefix on the dictionary and just append the `gen_` and `test_` prefix internally in the `generate_references.py` logic if you think is clearer.

About why not remove the layer, I'm not sure about what layer are you referring to, as the process you have described is very similar to the current one

1. The `generate_references.py` search, not for the inputs, but for the file and generator function in that file.
2.  We then generate the outputs after having imported our generator function.
3. Outputs are serialized, with inputs, in the references `.yaml` file.
4. The test_data (inputs + expected_outputs) are then passed to each test to compare against our estimators.

Maybe I'm not understanding what do you mean?

tests/references/SDFqEstimator/gen_sdfq.sage
Dioprz: So, something like this?

```python
def general_test(test_single_case, test_name, test_data):
    inputs_with_expected_outputs = test_data(test_name)
    map(test_single_case, inputs_with_expected_outputs)

# ...
def test_single_case_sdfq_lee_brickell(input_with_exp_output, epsilon):
    """
    Single test for the LeeBrickell-SDFq problem with an error tolerance of epsilon.
    """
    input, expected_complexity = input_with_exp_output
    n, k, w, q = input
    p = 2

    actual_complexity = LeeBrickell(
        SDFqProblem(n, k, w, q, **PARAMS), **PARAMS
    ).time_complexity(p=p)

    assert abs(actual_complexity - expected_complexity) < epsilon

general_test(test_single_case_sdfq_lee_brickell, "test_sdfq_lee_brickell", test_data)
```

Instead of this

```python

def test_sdfq_lee_brickell(test_data):
    """
    Test for the LeeBrickell-SDFq problem with an error tolerance of epsilon.
    """
    inputs_with_expected_outputs = test_data()
    epsilon = 0.01

    def test_single_case(input_with_exp_output):
        input, expected_complexity = input_with_exp_output
        n, k, w, q = input
        p = 2

        actual_complexity = LeeBrickell(
            SDFqProblem(n, k, w, q, **PARAMS), **PARAMS
        ).time_complexity(p=p)

        assert abs(actual_complexity - expected_complexity) < epsilon

    map(test_single_case, inputs_with_expected_outputs)
```

While that removes the code duplication, we now have some more troubles:

1. We have darkened the test process with another intermediary function.
2. We have splitted the information related to one test, into two different blocks (the one that defines the single_case_test, and the one that calls the evaluation of that `single_case_test`, via the `general_test_function`).
3. The argument `test_data` used in the *definition* of the functions is automatically provided by python fixtures when you execute the `pytest` command. I'm not sure that parameter can be "passed" into a function call (I can search if you want).

The biggest drawback to me is the number 2. We are saving nothing more than one line of code per test, because Python doesn't support multiline lambdas (to define the `test_single_case` directly in the `general_test` function call, for example), in exchange for a more convoluted test definition process.

Dioprz:
> Generally, there should be make commands not relying on docker for those having sage installed locally, i.e. make generate-tests-references etc.

I think, at least while we end the transition process and have some more free time to polish the path handling logic, we have to use only a docker version of this command. As you can see in `constants.py`, `Makefile`, and `generate_references.py`, the current logic is very dependant on the docker container file system structure.

This is not something to worry about, as I can implement the non-docker version of the command to require an argument with the path to the library, so everything will work as expected, but right now is maybe worthwile to follow with the transition to end as quick as possible. I can create an issue to not forget this. What are your thoughts on this?

Memphisd (changes requested):
Overall the new framework looks good. Just wondering if we can maybe lighten it a bit by e.g. avoiding some code duplications.

Once discussed (and in case changed) the PR is good to be merged.

tests/references/generate_references.py
Memphisd: I am just talking about the structure of the dictionary that we are generating. I don't think the dictionary should represent the code logic. The dictionary should just hold the test case data in the most intuitive form, which for me would be something like dict["Estimator"]["Algorithm"]["data"]=[(input1,output1),...] (or something similar).

tests/references/SDFqEstimator/gen_sdfq.sage
Memphisd: As I see it you could even automatize the call to

```general_test(...)```

similar to how we add the algorithm objects as attributes to an estimator object (in the estimator constructor).

All logic for the test the developer ever has to deal with would then be in the function ```test_sdfq_lee_brickell```.
Point 3 is not very clear to me yet.

Comment (FloydZ) [reply tests/references/SDFqEstimator/gen_sdfq.sage:63]: I'm in favor of keeping it as it is. We are talking about 100 lines of code. I think its fine if we loose a few lines here, but keep the whole thing readable, even for someone new to the project.

Comment (Dioprz) [reply tests/references/SDFqEstimator/gen_sdfq.sage:63]: @Memphisd 

> All logic for the test the developer ever has to deal with would then be in the function test_sdfq_lee_brickell.
Point 3 is not very clear to me yet.


1. The thing with pytest is... It collects only *function definitions* starting with the `test_` prefix, but it doesn't execute the file like a typical `python3 file.py` run does. That makes your approach based on *function calls* instead of *function definitions* hard to implement. And while I'm sure what you want is doable, it could involve a lot of work overwriting the collection method that pytest uses by default. Another option could be to create a big test, with each subtest as a inner definition function:

```python
def test_sdfq(test_data):
  def general_test(test_single_case, test_name, test_data):
    pass

  def sdfq_lee_brickell(input_with_exp_output):
     pass

  general_test(sdfq_lee_brickell, "sdfq_lee_brickell", test_data):
     pass
  # And repeat this with each test
```

While I could expect this to work, I still don't see the gain. We are obscuring the testing process by adding another intermediary function in exchange for 1 or 2 lines of code per-test.

2. The test developers *always* should have in mind that  they need to deal with three files, the `gen_estimator` file, the inputs dictionary file, and the `test_`file. They can't be understood as unrelated things, because each one represent a different step in the testing flow.

3. And in my previous last point, about pytest fixtures, what I want to say is: the parameter `test_data` I'm giving to each test, is provided automatically by a pytest feature called `fixture`. That fixture is configured in the `conftest.py` file, and it's because of this that we will load the reference `.yaml` only one time on each `pytest` command execution (even if we are running all the estimators tests). 
Because it is provided directly in the *function definition* and **not** in *function calls*, if we want to achieve something like what you are purposing, I should need to go really deep in the pytest module to understand how to capture that information and then pass it as an argument for the general testing function call.

While I have the same concern as you about the code duplication, what I found is that any alternative I can find or think of, involves either obscuring the testing process, or using advanced pytest testing features/hacky techniques. So I prioritize the maintanability of the code by using the most straightforward implementation I can, even if it involves those extra lines.

tests/references/generate_references.py
Dioprz: I, however, think that the dictionary structure should signalize as much as it can about where the information is being used. If we eliminate that prefix from the `REFERENCES_INPUT` dictionary, what prevents the test developer (at the code level, because of course the documentation is clear on that) to think that those inputs will be passed directly to the `test` writen in `validations`? The prefixes are helping to understand, on each step of the data flowing, what is the purpose of that data and who will consume it.

tests/helper.py
```diff
@@ -0,0 +1,3 @@
+DOCKER_YAML_REFERENCE_PATH = (
+    "/home/cryptographic_estimators/tests/validations/reference_values.yaml"
```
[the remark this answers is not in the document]
Dioprz: Please see https://github.com/Crypto-TII/CryptographicEstimators/pull/161#issuecomment-2244014711 for the explanation of why can't be relative at the moment (but will be changed once we finish the migration).

docs/CONTRIBUTING.md
```diff
@@ -1,46 +1,61 @@
...
 refactor/modify_base_problem
 feature/implement_dummy_estimator
 fix/algorithm_parameter
 ```
 
 ### Pull request
+
 1. Only create pull requests to the `develop` branch.
-2. Fulfill the template
+1. Fulfill the template
```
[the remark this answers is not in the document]
Dioprz: Sorry, this was produced by my [formatter ](https://mdformat.readthedocs.io/en/stable/users/style.html#ordered-lists). Now I made the changes and configured it properly so this will not happen again.

docs/tests.md
```diff
@@ -0,0 +1,345 @@
...
+or all tests execute
+
+```sh
+make testall
+```
+
+or if you have Apple Silicon M1 Chip
+
+```sh
+make test-m1
```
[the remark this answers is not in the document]
Dioprz: Done

docs/tests.md
```diff
@@ -0,0 +1,345 @@
...
+     and specifies the inputs for the generator functions.
+   - To add a new estimator and its inputs, follow this structure within the
+     dictionary:
+
+   ```python
+   REFERENCES_INPUTS = {
+       "<estimator_name>Estimator": {
+           "gen_<estimator_name>": {
+               "gen_<estimator_name>_<function_name_1>": {
+                   "inputs": [(256, 128, 64, 251), (961, 771, 48, 31)]
```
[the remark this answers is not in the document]
Dioprz: That key is later renamed as `inputs_with_expected_outputs` in the `reference_values.yaml` file for clarity. Do you think it is better to remove it even so?

tests/references/SDFqEstimator/gen_sdfq.sage
Memphisd: > I'm in favor of keeping it as it is. We are talking about 100 lines of code. I think its fine if we loose a few lines here, but keep the whole thing readable, even for someone new to the project.

I think it is not only about code duplication here but about how to write the tests. I think the nested function structure overcomplicates it. All tests for all functions follow basically the same principle, its a KAT test with (in some cases) a certain tolerance. However the current framework requires you to write the KAT test logic, that is read the inputs and outputs create the objects, check against output with certain tolerance, for every test. I am questioning if we can do better here.
I think it is worth spending a few thoughts on it now. Once implemented, it will be much more work should we at some point decide to change it. However, if after careful consideration you still say there are too many technical obstacles in automatizing the logic, I guess we can proceed.

tests/references/generate_references.py
Memphisd: Maybe I am misunderstanding the whole framework.
From my understanding there should be two things:
1. Functions to generate the dictionary for each estimator
2. Test functions that run KAT tests against the generated dictionary

But from what you write there is more than that? I suggest we clarify this on a conceptional level first, then talk about the technicalities.

tests/references/SDFqEstimator/gen_sdfq.sage
Javierverbel: @Dioprz is it clear what @Memphisd suggestion here is? If so, did you have time to think about a solution?

tests/references/generate_references.py
Dioprz: @Memphisd Sorry for the delay.

  Thinking about your last response, I realize you enlightened a critical point: my previous documentation was not good. We need to ensure that any contributor understands the whys and hows of the new framework by just reading the documentation, and that's clearly not happening.

  I have made a complete revamp of the documentation to fix this, so please review the latest commit at your convenience, and to ensure clarity, let's continue this discussion in a new thread.

  Lastly, I would appreciate more comprehensive descriptions of your concerns/suggestions in the future. I had spent a considerable amount of time writing detailed responses to your comments (and probably you do it too), just to find two weeks later that we, sadly, were not even in the same page of the discussion. I really appreciate your feedback, and it would be great if we could understand each other faster next time!

Dioprz:
Thanks to the discussion with @memphisd, I realized my previous documentation needed significant improvements. Please take a look at the changes [here](https://github.com/Crypto-TII/CryptographicEstimators/blob/80d942c5464ad0987a9ba44e512096a67bbcdd38/docs/tests.md) (make sure to view the *rich/preview diff* instead of the *source/code diff*, so you can also see the diagram I added):

To ensure clarity moving forward, I propose we use previous discussions as a reference point, but start fresh with new threads/reviews to address any remaining concerns based on the updated documentation.

Sorry for the additional review request, @Javierverbel  and @FloydZ , but as @memphisd pointed out, this is an area that needs further attention.

tests/references/SDFqEstimator/gen_sdfq.sage
Dioprz: No, not really. I can't find any better solution to what I proposed.

My line of thinking is like this:

1. We have some redundant lines of code, corresponding to the KAT tests logic.
2. If we want to remove them, we need to abstract those lines with either a class or a function.
3. A class is completely overkilled for this purposes.
4. A function that abstract those lines, must then be called with the non-redundant information as parameters, i.e.:
    1. The function that evaluates inputs to produce our estimators outputs.
    2. The deserialized dictionary.
    3. The function name, so we can access the correspondant data (inputs/exp_outputs) in the deserialized dictionary. (I think this could be abstracted too... Maybe I can infer it from the function name if they match... But anyways).
5. Because pytest ONLY collects test *definitions* starting by the `test_` prefix, the function call must be inside a bigger function starting with such prefix.
6. That bigger function must contain all the non-redudant information for each test, and then call the general function on that information.

This is what led me to propose the code I proposed in my previous response. And there I mention what drawbacks I see in such solution.

Maybe I'm making an incorrect deduction in some step, that can be solved in other, more efficient way, but I can't see it.

And, just for the record, I`m sure this lines could be abstracted in a neat way from a test-developer perspective by using advanced pytest features (like implementing our custom pytest collector function, or search if pytest could trigger the call for functions from a centralized file, for example). But this will take me a considerable amount of time to learn how to do it, and then implement it; not to mention that it will result weird to someone already familiar with pytest, as we are changing the game rules for the tool (pytest).

In case it helps in some way, the nested function structure can be removed. I like to think with maps and data transformations, but the logic can be translated into a for-loop for a more imperative approach if that sound better to you:

```python
def test_sdfq_lee_brickell(test_data):
    """
    Test for the LeeBrickell-SDFq problem with an error tolerance of epsilon.
    """
    inputs_with_expected_outputs = test_data()
    epsilon = 0.01

    def test_single_case(input_with_exp_output):
        input, expected_complexity = input_with_exp_output
        n, k, w, q = input
        p = 2

        actual_complexity = LeeBrickell(
            SDFqProblem(n, k, w, q, **PARAMS), **PARAMS
        ).time_complexity(p=p)

        assert abs(actual_complexity - expected_complexity) < epsilon

    map(test_single_case, inputs_with_expected_outputs)
```

Becomes

```python
def test_sdfq_lee_brickell(test_data):
    """
    Test for the LeeBrickell-SDFq problem with an error tolerance of epsilon.
    """
    inputs_with_expected_outputs = test_data()
    epsilon = 0.01

    for input, expected_complexity in inputs_with_expected_outputs:
        input, expected_complexity = input_with_exp_output
        n, k, w, q = input
        p = 2

        actual_complexity = LeeBrickell(
            SDFqProblem(n, k, w, q, **PARAMS), **PARAMS
        ).time_complexity(p=p)

        assert abs(actual_complexity - expected_complexity) < epsilon
```

docs/tests.md
```diff
@@ -0,0 +1,415 @@
...
+
+## Testing Suite Architecture Overview
+
+This section provides an overview of the testing architecture for our Python
+library, which employs two types of tests: **regression tests** and
+**doctests**. Both types are powered by pytest, with doctests **(WIP)** being
+specifically based on the Python's doctest module.
+
+## Regression Tests: Understanding the Process
+
```
Javierverbel, line 11: I would define this tests differently to highlight that the estimates compute by the estimators library somehow match those made by state of the art of estimators. 

I would suggest to include something like this:
"The goal of these tests is to ensure that the complexity estimates made by the `CryptographicEstimators` library match the ones produced by state-of-the-art estimators implemented by researchers of the cryptographic community."

docs/tests.md
```diff
@@ -0,0 +1,415 @@
...
+
+    gf --> preeo;
+    inputs --> preeo;
+    preeo --> ser;
+    ser -- input --> estimator;
+    estimator -- output --> oc;
+    ser -- expected output --> oc;
+
+```
```
Javierverbel, line 44: If possible, it would be good to add the dictionary name `REFERENCES_INPUTS` in the figure

docs/tests.md
```diff
@@ -0,0 +1,415 @@
...
+### Writing Tests: A Step-by-Step Guide
+
+This guide will walk you through the process of adding a new regression test to
+our Python library. We'll assume you're already familiar with Python, pytest,
+and the basic structure of our estimators.
+
+#### 1. Create Your Generator Function
+
+- Create a new Python file (`gen_<your_new_estimator>.py`) within the
+  `tests/references/<YourNewEstimator>Estimator/` directory.
```
Javierverbel, line 76: After each  `tests/references/<YourNewEstimator>Estimator/` it would be good to add an example maybe it the SDEstimator or the DummyEstimator

docs/tests.md
```diff
@@ -0,0 +1,415 @@
...
+- Define your generator function that will produce expected outputs based on the
+  provided inputs. Adhere to the naming convention
+  (`gen_<your_new_estimator>_<test_function_name>`)
```
Javierverbel, line 79: add example.

docs/tests.md
```diff
@@ -0,0 +1,415 @@
...
+    "<YourNewEstimator>Estimator": {
+        "gen_<your_new_estimator>": {
+            "gen_<your_new_estimator>_<your_test_function_name>": {
+                "inputs": [(input_value_1, input_value_2, ...),  (...), ...] # Your input tuples here
+            },
+            # Add more functions as needed...
+          },
+    },
+    # ... more estimators...
+}
```
Javierverbel, line 114: Add example.

Javierverbel (dismissed):
From my side, after some small editorial comments it can be merged.

tests/references/SDFqEstimator/gen_sdfq.sage
FloydZ: I think, we have something more like this in mind:
```python
# NOTE: the missing `test_` prefix, s.t. the function is ignored by the pytest framework
# This would be in `tests/validations/test_sdfq.py`
def single_case_leebrickell(input): 
    n, k, w, q = input 
    return LeeBrickell(SDFqProblem(n, k, w, q)).time_complexity(p=2)

# This would be in `tests/validations/test_sdfq.py`
def single_case_stern(input): 
    n, k, w, q = input 
    return Stern(SDFqProblem(n, k, w, q)).time_complexity()

# NOTE: this function would still be in conftest.py
def test_data(yaml_references: dict):
    """
    Fixture to retrieve test data from YAML references.
    """

    def get_test_data(test_file: str = None, test_function: str = None):
        """
        Get inputs and expected outputs for a specific test function.
        If test_file or test_function is not provided, it will be inferred from the caller.
        """
        if test_file is None or test_function is None:
            caller_frame = inspect.currentframe().f_back
            if test_file is None:
                test_file = os.path.splitext(
                    os.path.basename(caller_frame.f_code.co_filename)
                )[0]
            if test_function is None:
                test_function = caller_frame.f_code.co_name

        expected_complexity, input_data, epsilon = yaml_references[test_file][test_function]
        # admittedly: this is shady, as you need catch the case that multiple test cases are
        # mapped to the same algorithm, like `gen_stern` and `gen_stern_range` in our sdfq example.
        actual_complexity = test_function(*input_data)
        assert abs(actual_complexity - expected_complexity) < epsilon
```
The idea is to reduce the usage of `pytest` to basically a single function: `test_data()`, which does everything. From the test-case extraction to the actual complexity comparison. 

This has the big advantage that the work a new contributor would have to do, is minimal. Basically write a function for each algorithm the new estimator contains (in this example only `single_case_leebrickell` and `single_case_stern`)

@Dioprz  please let me know whether this is technically feasible at all. Specially this mapping of multiple testcase to a single test function (in our example sdfq_stern).

PR #165, opened 2024-07-30
Feat/add bike estimator sonar

AnaCaceres:
### Description
- (purpose of the PR, how did you implement it)

### Review process
- (explain how to review it)

### Pre-approval checklist


Dioprz (approved):
As before, it looks good to me.

Out of curiosity, are there anything we can do to avoid the code duplication without the sonar exception? Is this something that could happen again with new estimators, or just an isolated case?

AnaCaceres:
For duplicity errors there is no other option than to skip all sonar analysis for the folder cryptographic_estimators/BIKEEstimator/BIKEAlgorithms, as this is the only way for sonar not to analyze the duplicity. 

This might happen in the future if there is a new estimator created that is practically the same than another existing one. In this case Sonar will detect the duplicity. The only way of not having to disable the sonar analysis is handling the duplicated code itself.

FloydZ (approved):

Javierverbel (approved):

Issue #162, opened 2024-07-15
Refactor Docstring Style to Google Style


Dioprz:
Currently, our codebase uses a non-standard docstring style inspired by [`sage docstrings`](https://doc.sagemath.org/html/en/developer/coding_basics.html#documentation-strings). While functional, this style presents several drawbacks, including poor support for automatic documentation generation tools like Sphinx and inconsistencies in formatting that includes:

  - `INPUT:` and `OUTPUT:` need one colon, but `EXAMPLES::` and `TESTS::` require
    two.
  - Some blocks like `.. MATH::`, `.. NOTE::`, require atypical leading dots.
  - The parameter type declaration (if declared) has a very particular way to be
    declared and separate it from the parameter description:
    ```python
    """
    - ``p`` -- prime integer (default: `2`); coprime with ``n``
    """
    ```

This issue proposes migrating to the industry-standard **Google docstring style** to address these shortcomings while capitalizing on the ongoing migration from `sage doctest` to Python's native `doctest`.

**Current Style:**

```python
"""
Construct an instance of crossbred estimator

The Crossbred is an algorithm to solve the MQ problem [JV18]_...
.. NOTE::

    Our complexity estimates are a generalization over any field of size `q` of the complexity formulas given in
    [Dua20]_, which are given either for `q=2` or generic fields.

INPUT:

- ``problem`` -- MQProblem object including all necessary parameters
- ``h`` -- external hybridization parameter (default: 0)
- ``w`` -- linear algebra constant (2 <= w <= 3) (default: 2.81)
- ``max_D`` -- upper bound to the parameter D (default: 20)
- ``memory_access`` -- specifies the memory access cost model (default: 0, choices: 0 - constant, 1 - logarithmic, 2 - square-root, 3 - cube-root or deploy custom function which takes as input the logarithm of the total memory usage)
- ``complexity_type`` -- complexity type to consider (0: estimate, 1: tilde O complexity, default: 0)

EXAMPLES::

    sage: from cryptographic_estimators.MQEstimator.MQAlgorithms.crossbred import Crossbred
    sage: from cryptographic_estimators.MQEstimator.mq_problem import MQProblem
    sage: E = Crossbred(MQProblem(n=10, m=12, q=5))
    sage: E
    Crossbred estimator for the MQ problem with 10 variables and 12 polynomials

"""
```

**Proposed Style (Google Docstring):**

```python
  """Construct an instance of crossbred estimator.

[...]

PR #169, opened 2024-08-12
Refactor/new docstrings styling

Dioprz:
### Description

After three weeks researching about alternatives to run doctests dinamically, I finally found how to put the pieces together. The result is this PR where we have:

- The implementation of the first doctests migration from Sagemath to Python.
- The changes proposed here https://github.com/Crypto-TII/CryptographicEstimators/issues/162, to change the docstrings style into a more standard approach (Google docstrings).
- The creation of a pytest CLI option, so we can trigger the dynamic definition of a boolean flag via pytest fixtures.
-  A new docstrings logic to dinamically decide if we run or not long doctests.

I made the changes in just one file so we can discuss any aspect, before go all-in.

#### How this changes work

At `conftest.py`
1. We now have a [pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html) called automatically on each `pytest` CLI execution, as well as a new pytest CLI option called `--long-doctests`,  that defaults to False but changes to True when provided.
2. Whatever value we get from this CLI option, we put it into a `long_doctests` variable that is later passed to the *"context"* of our doctests via [`doctest_namespaces` fixtures
](https://docs.pytest.org/en/7.1.x/how-to/doctest.html#doctest-namespace-fixture)

At `sd_estimator.py`

We now have to collect all the tests that could take a long time to run, and put them at the end of the `Tests:` section. This way, we can use the `long_doctests` flag comming from the pytest context, and with the conditional
```python 
>>> if not long_doctests:
 ...     pytest.skip()
```
Ensure that every test block after that will be run only if the flag `long_doctests` is True (i.e., if we made the `pytest` call with the `--long-doctests` option).

#### Pros and drawbacks

**Pros**
1. The `#long time` functionality of Sage markers is, essentially, preserved.

**Drawbacks**
1. Now we don't have the granularity of a docstring marker, so we can't alternate between long and short time blocks inside a docstring. Every long test should be at the end of the `Tests` section.
2. When the `--long-doctests` option is not provided to the `pytest` call, all the tests containing long test are reported as **skipped** instead of **passed**. While this shouldn't have any technical implication, because short failing tests still produces a pytest error, I think it's worth to mention it.

### Review process
- For short tests/examples:
    - Run `pytest --doctest-modules cryptographic_estimators/SDEstimator/sd_estimator.py` in the shell. It will work, and the pytest return will be `1 skipped`.
    - Change some value of the expected output for the `Examples:` section of the docstring, and save the file. Run the same command again and, as expected, pytest will return `1 failed`.
- For long tests:
  - Run `pytest --long-doctests --doctest-modules cryptographic_estimators/SDEstimator/sd_estimator.py`. Long tests defined after the previousd conditional, will be executed and the pytest call will return `1 passed`.

### Pre-approval checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -40,5 +40,9 @@ jobs:
           rm -rf /tmp/.buildx-cache
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
 
-      - name: Run tests
+      - name: Run Sage doctests
         run: make docker-test
+
+      - name: Run pytest
+        run: make docker-pytest
+
--- a/Makefile
+++ b/Makefile
@@ -71,10 +71,67 @@ docker-doc: docker-build
 	@make mount-volume-and-run && make generate-documentation && make stop-container-and-remove container_name="container-for-docs"
 
 docker-test: docker-build
-	@docker run --name container-for-test -d -it ${image_name} sh && docker exec container-for-test sage -t --long -T 3600 --force-lib cryptographic_estimators && docker stop container-for-test && docker rm container-for-test
+	@echo "Removing previous container...."
+	@make stop-container-and-remove container_name="container-for-test" \
+		|| true
+	@echo "Creating container..."
+	@docker run --name container-for-test -d -it ${image_name} sh \
+		&& docker exec container-for-test sh -c " \
+		sage -t --long --timeout 3600 --force-lib \
+		# cryptographic_estimators/SDEstimator/ \
+		cryptographic_estimators/DummyEstimator/ \
+		cryptographic_estimators/LEEstimator/ \
+		cryptographic_estimators/MAYOEstimator/ \
+		cryptographic_estimators/MQEstimator/ \
+		cryptographic_estimators/MREstimator/ \
+		cryptographic_estimators/PEEstimator/ \
+		cryptographic_estimators/PKEstimator/ \
+		cryptographic_estimators/RegSDEstimator/ \
+		cryptographic_estimators/SDFqEstimator/ \
+		cryptographic_estimators/UOVEstimator/ \
+		cryptographic_estimators/base_algorithm.py \
+		cryptographic_estimators/base_constants.py \
+		cryptographic_estimators/base_estimator.py \
+		cryptographic_estimators/base_problem.py \
+		cryptographic_estimators/estimation_renderer.py \
+		cryptographic_estimators/helper.py \
+		" \
+		&& echo "All tests passed." \
+		|| echo "Some test have failed, please see previous lines."
+	@echo "Cleaning container..."
+	@make stop-container-and-remove container_name="container-for-test"
+
 
 docker-testfast: docker-build
-	@docker run --name container-for-test -d -it ${image_name} sh && docker exec container-for-test sage -t cryptographic_estimators && make stop-container-and-remove container_name="container-for-test"
+	@echo "Removing previous container...."
+	@make stop-container-and-remove container_name="container-for-test" \
+		|| true
+	@echo "Creating container..."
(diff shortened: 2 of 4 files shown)


Dioprz:
I made one change to the name of the flag from `long-doctests` to `skip-long-doctests`, so it is easier to understand. The flag usage is inverted because of this:

  ```python

  ### Previous
  ### >>> if not long_doctests:
  ### ...     pytest.skip()

  ### New
  >>> if skip_long_doctests:
  ...     pytest.skip()

  ```

  And the testing process looks now like this

  1. `make docker-build`
  2. `make docker-run`
  3. `pytest --doctest-modules cryptographic_estimators/SDEstimator/sd_estimator.py` -> Produces as output `1 passed`
  4. `pytest --skip-long-doctests --doctest-modules cryptographic_estimators/SDEstimator/sd_estimator.py` -> Produces as output `1 skipped`
  5. You can change any value in the `Examples:` block to test the short execution failure (remember to rebuild the docker with your change).
  6. You can change any value in the `Examples:` or `Tests:` blocks to test the long execution failure.

  It would be great if this PR could have at least three approvals as it redefines a new standard for the library. Please let me know if there are any objections.

Dioprz:
As shown in ad0b9d1, a problem emerges with the CI when changing the docstring style.

**The problem:** Sage collects not only doctests with the Sage format, but *also* doctests with the Python format. Because the new format uses a feature that we only want to execute with Pytest and pure Python (the `--skip-long-doctests`, that feeds the `skip_long_doctests` fixture used in the new test style), the doctest fails.

**The solution idea:**

We need to be able to:

1. Introduce the style changes in the doctests of some file *X*.
2. Exclude *X* from being collected by the Sage doctest run process.
3. Include *X* into the collection of Pytest.

This way, we will make a safe testing transition in each PR, as in any state of the repo, every doctest will be evaluated by either Sage or Python, but not both.

**Solution implementation:**

- At [5f2b143](https://github.com/Crypto-TII/CryptographicEstimators/pull/169/commits/5f2b1438ed2abca37ed086958ddff817490a3584), I made two changes:
  1. Create a new command `docker-pytest`[^1]. This command executes Pytest with pure Python, and declares every subdirectory and file inside `cryptographic_estimators/`, plus the `tests/` directory.
  2. Refactor the commands `docker-test` and `docker-testfast` so they include every subdirectory inside `cryptographic_estimators/`.

  While the three commands declare the same subdirectories, being extremely verbose, you can see too that each line commented in `docker-test[fast]` is not commented in `docker-pytest`. The intention behind this is to have the next workflow during the docstring transition:
  1. We will have just one docstring-migration PR per estimator. Each one of those PRs will apply the new style to *every* file inside the estimator subdirectory.
  2. We comment the estimator directory in `docker-test` and `docker-testfast`, and uncomment it in `docker-pytest`.
  3. The transition is done for that estimator, and we can move forward with the next one.
  4. At the end, the `docker-test[fast]` command will disappear, and we can remove the explicit directory listing from the pytest command.

  [^1]: We already had one `docker-pytest` command declared. However, and as you can see in the comment of 5f2b143, such command wasn't being used in **any** GH action (you can confirm with `grep -r "docker-pytest" .github`). For that reason, and because that command wasn't correctly structured, I decided to take its name and re-implement it.

- At [238141b](https://github.com/Crypto-TII/CryptographicEstimators/pull/169/commits/238141bd02ee6a8cd138b18c6cef94c7568f169b), I include the `make docker-pytest` command into our CI.

FloydZ (approved):
Looks very good.
Cool approach for splitting the tests for `sage` and `python`.

Javierverbel (approved):

Memphisd (approved):
Looks good to me.

PR #170, opened 2024-08-12
Refactor/avoiding large integers booleansolve fxl

Javierverbel:
### Description
- Avoid large integers operations in `booleansolve_fxl.py`. It was given some errors for the python tests.

### Review process
-  Verify that the pytests are executed without errors


Dioprz:
The previous problematic test:

```python
>>> from cryptographic_estimators.UOVEstimator import UOVEstimator
>>> A = UOVEstimator(n=184, m=72, q=256, theta=None)
>>> A.table(show_all_parameters=1) # long time
```

Works now with pure python, and the worst case scenario that we have (`A = UOVEstimator(n=244, m=96, q=256, theta=None)`, does it too.

Thanks for the fix!

Dioprz (approved):

PR #174, opened 2024-09-12
Fix/value error uov mayo

fmerino21:
### Description
- Minor fix to avoid a value error when `n = 1.5 * m` in UOV `intersection_attack`
- Check if `time < 0` in `exhaustive_search` to prevent negative values in MAYO `direct_attack`
Specific example: `n=45, m=30, q=16, k=9, 0=8`

### Review process
- Check the implementation of the fixes  

### Pre-approval checklist


Javierverbel:
The method `expected_number_solutions` should be removed from the `mayo_problem.py` this method doesn't make sense for the estimators of schemes. Same comment for the attribute `nsolution`.

Instead of raising an error when `time <0`. We should remove the method and attribute that regard to the number of solutions.
Then, we call the build the instance of the `MQEstimator` inside the `direct_attack` without parsing the argument `nsolutions`

Javierverbel (approved):

PR #176, opened 2024-09-16
Refactor/python docstrings for sd estimator

Dioprz:
### Description

Implementation of https://github.com/Crypto-TII/CryptographicEstimators/pull/169 for SDEstimator.

Formatting changes were made by [Black](https://github.com/psf/black).

### Review process

- A quick glance at the new docstrings would be great.

**Estimator-specific doctests**
```shell
make docker-build
make docker-run
pytest --doctest-modules -n auto -vv cryptographic_estimators/SDEstimator/
```

**Cumulative test (with all the already migrated docstrings)**
```shell
make docker-pytest
```

### Pre-approval checklist


Dioprz (commented):
The Sonarcloud test failing is quite... Dumb.

Basically, it complains because Black has improved the formatting of some files by splitting their content into multiple lines. This means that if we previously had one line of unreadable code duplicated, and it's now split into three lines of readable code, it complains because there are "too many duplicated lines of code". This feels incorrect, in my opinion.

Anyway, I can offer you two options:

1. Ignore the Sonarcloud warning in this PR and future docstring-migration-related PRs, knowing that some of those checks could be sensible and reasonable, and address them later.

2. Ignore the Sonarcloud warning in this PR, and disable Black for the incoming PRs so Sonarcloud doesn't complain on the next ones. (Very discouraged from my side, as we are losing a great opportunity to improve the code quality by making it standard on formatting. However, it can also help in case you feel tihs PR too noisy).

Both options include ignoring the Sonarcloud warning for this PR, as I don't have a quick way to undo the formatting changes, so basically it would imply making all the docstring changes from scratch.

Here I'm pointing out some examples of the changes introduced by Black.

cryptographic_estimators/SDEstimator/SDAlgorithms/may_ozerov.py
```diff
@@ -382,100 +395,122 @@ def initialize_parameter_ranges(self):
...
+                    for p1 in range(
+                        max(new_ranges["p1"]["min"], (p2 + 1) // 2),
+                        min(new_ranges["p1"]["max"], k1 - p2 // 2),
+                    ):
+                        indices = {
+                            "p": p,
+                            "p1": p1,
+                            "p2": p2,
+                            "l": l,
+                            "r": self._optimal_parameters["r"],
```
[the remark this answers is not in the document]
Memphisd: Hey Diego,

I agree that sonar is a bit annoying with these things. But here I think we should discuss the formatting. I don't think that in the particular case of 

```
New code
            {"type": "ineq", "fun": self._inject_vars(lambda x: self.rate(x) - x.p)},
            {
                "type": "ineq",
                "fun": self._inject_vars(
                    lambda x: (1.0 - self.rate(x) - x.l) - (self.w(x) - x.p)
                ),
            },
            {"type": "ineq", "fun": self._inject_vars(lambda x: self.w(x) - x.p)},
```

```
Old code
            {"type": "ineq", "fun": self._inject_vars(lambda x: self.rate(x) - x.p)},
            {"type": "ineq", "fun": self._inject_vars(lambda x: (1.0 - self.rate(x) - x.l) - (self.w(x) - x.p)),},
            {"type": "ineq", "fun": self._inject_vars(lambda x: self.w(x) - x.p)},
```

The code becomes better readable. For all constraint definitions this basically holds as the start of the line is the same (or very slightly different), which allows to focus on the last part. For the new version is is not obvious and has to be checked manually. I suggest reverting the changes with respect to the constraint definitions and ignore the line length limitation for them.

Also it seems you are using 79 chars line length. I think we are all using the PyCharm IDE which defaults at 120 char per line. I know that PEP recommends 79, but this is also historical and often nowadays longer lines are common. Personally I prefer longer lines, the exact width could be up for debate but I think it should be longer than 79. Depending on the outcome on this discussion, some of the changes should be reverted as well, which then should hopefully resolve the sonar complaints.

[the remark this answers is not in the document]
Dioprz: Thanks for pointing that out.

I completely agree with you on that modification. As you mentioned, is consequence of the line-length used by default by Black.

About the line-length, I don't have any strong opinion on that. Whatever works for you, works for me.

I have applied the reformatting to all the estimator files with a 120 line-length. Let's see if that resolves the Sonarcloud complaint.

Dioprz:
@Javierverbel Thank you so much for fixing the Sonarcloud warning 💯 .

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -30,29 +38,25 @@
 
 class BJMM(SDAlgorithm):
     def __init__(self, problem: SDProblem, **kwargs):
-        """
-        Complexity estimate of BJMM algorithm in depth 2,3
+        """Complexity estimate of BJMM algorithm in depth 2,3.
 
-        The algorithm was introduced in [BJMM12]_  as an extension of [MMT11]_.
+        The algorithm was introduced in [BJMM12] as an extension of [MMT11].
```
FloydZ, line 43: weren't these underscores `[BJMM]_` important for the automatic generation of the citations?

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -169,49 +159,41 @@ def get_optimal_parameters_dict(self):
         return a
 
     def __repr__(self):
-        """
-        """
+        """ """
```
FloydZ, line 162: just remove it

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -169,49 +159,41 @@ def get_optimal_parameters_dict(self):
...
+        The expected weight distribution is as follows:
 
             +--------------------------+-------------------+-------------------+
             | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
             |           w - 2p         |        p          |        p          |
             +--------------------------+-------------------+-------------------+
 
-        INPUT:
+        Args:
+            problem (SDProblem): SDProblem object including all necessary parameters.
```
FloydZ, line 181: is this still a valid way to describe the arguments for the automatic html docs generation?

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -313,76 +295,72 @@ def _time_and_memory_complexity(self, parameters: dict, verbose_information=None
...
-        expected weight distribution::
-
-            +--------------------------+-------------------+-------------------+
-            | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |        p          |        p          |
-            +--------------------------+-------------------+-------------------+
+        """Complexity estimate of BJMM algorithm in depth 2.
 
-        INPUT:
+        The algorithm was introduced in [BJMM12] as an extension of [MMT11].
```
FloydZ, line 347: _

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_dw.py
```diff
@@ -19,56 +19,61 @@
...
+        Args:
+            problem (SDProblem): Syndrome decoding problem instance.
+            k (int): Dimension of the code.
+            w (int): Hamming weight of error vector.
+            mem (int, optional): Upper bound on the available memory (as log2), default unlimited.
+            hmap (bool, optional): Indicates if hashmap is being used (default: True).
+            p_range (list, optional): Interval in which the parameter p is searched (default: [0, 25], helps speeding up computation).
+            memory_access (int, optional): Specifies the memory access cost model (default: 0, choices: 0 - constant, 1 - logarithmic, 2 - square-root, 3 - cube-root or deploy custom function which takes as input the logarithm of the total memory usage).
+
+        References:
```
FloydZ, line 60: ok here we change the "citation type" from `[MMT]` to this. Is there a reason for this? Because I think this way leads to 
1 ) alot of copy paste
2) errors because of that

I liked the first central way of collecting citations, but maybe there are reasons against it.

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_dw.py
```diff
@@ -19,56 +19,61 @@
...
+            .. [EB22] Enguehard, Élise, and Anja Becker. "Quantum attacks on McEliece and Niederreiter cryptosystems." Designs, Codes and Cryptography 90.1 (2022): 1-23.
```
FloydZ, line 61: Btw this should be [Esser, Bellini]

cryptographic_estimators/SDEstimator/SDAlgorithms/dumer.py
```diff
@@ -29,39 +36,33 @@
 class Dumer(SDAlgorithm):
     def __init__(self, problem: SDProblem, **kwargs):
         """
-        Complexity estimate of Dumer's ISD algorithm
+        Complexity estimate of Dumer's ISD algorithm.
```
FloydZ, line 40: here the underscore is untouched

cryptographic_estimators/SDEstimator/SDAlgorithms/stern.py
```diff
@@ -28,24 +36,22 @@
 class Stern(SDAlgorithm):
     def __init__(self, problem: SDProblem, **kwargs):
         """
-        Construct an instance of Stern's estimator [Ste88]_, [BLP08]_.
+        Construct an instance of Stern's estimator.
```
FloydZ, line 39: citation completely deleted.

cryptographic_estimators/SDEstimator/sd_helper.py
```diff
@@ -20,92 +20,111 @@
...
-    [BLP08] Bernstein, D.J., Lange, T., Peters, C.: Attacking and defending the mceliece cryptosystem.
-    In: International Workshop on Post-Quantum Cryptography. pp. 31–46. Springer (2008)
+    Returns:
+        The complexity estimate of the Gaussian elimination routine.
 
-    EXAMPLES::
-
-        sage: from cryptographic_estimators.SDEstimator import _gaussian_elimination_complexity
-        sage: _gaussian_elimination_complexity(n=100,k=20,r=1) # random
+    References:
```
FloydZ, line 99: citation issue

cryptographic_estimators/SDEstimator/sd_problem.py
```diff
@@ -47,53 +44,56 @@ def __init__(self, n: int, k: int, w: int, **kwargs):
...
         self.parameters[SD_ERROR_WEIGHT] = w
 
-        self.nsolutions = kwargs.get("nsolutions", max(
-            self.expected_number_solutions(), 0))
+        self.nsolutions = kwargs.get("nsolutions", max(self.expected_number_solutions(), 0))
 
     def to_bitcomplexity_time(self, basic_operations: float):
         """
-        Returns the bit-complexity corresponding to basic_operations field additions
+        Calculates the bit-complexity corresponding to the number of field additions.
```
FloydZ, line 51: I think adding:
> , which are the `basic_operations` for SDAlgorithms

should be helpful. As we often refer to `basic_operations` in the help files

FloydZ (commented):
only small comments, mainly about the citations and if the html doc generation still works

cryptographic_estimators/SDEstimator/sd_helper.py
Dioprz: Which is the error, sorry? The changed format (year position, pages instead of pp., etc)?

cryptographic_estimators/SDEstimator/SDAlgorithms/stern.py
Dioprz: This is really bad.

Thanks for the catch.

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_dw.py
Dioprz: I have no clue why that could happen. I mean, the information to create that reference section doesn't even exist in the previous docstring.

AI was making some really weird things here.

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_dw.py
Dioprz: I completely removed the reference section, as it didn't even exist in the original docstring.

Dioprz:
> only small comments, mainly about the citations and if the html doc generation still works

Thank you for the careful review, @FloydZ.

Some things that are worth explaining:

- Sage uses reStructuredText-based docstrings, while Google-style docstrings don't (at least by default). This implies that:
  - The current HTML docs generation process will likely be incompatible with the new docstring style out of the box. I believe we will need, for example, the [Sphinx Napoleon extension](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#getting-started) to render our new docstring style.
  - Because I wasn't expecting retrocompatibility, and the main goal right now is to have a pure-python executable library, I wasn't taking too much care to preserve the reST format.
  - This is relatively easy to fix later. Due to the citation format, we only need a regex expression to look for places where the format `[XXXX]` is present and modify them as needed.
  
- Regarding the more serious errors like removing citations or creating new reference sections with incorrect information, these were the result of a very experimental prompt and workflow process on my side. Because I concentrated so much on not creating or modifying values for the `Examples:` or `Tests:` sections, I completely forgot to refine the process for the other sections of the docstrings.

- Once I sent this PR, I immediately improved the prompt used as well as my workflow process with the learnings made. Now I can inspect and review more carefully what is being generated by AI, so these errors shouldn't be present in subsequent PRs.

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
Memphisd: yes, I am not sure it was actually functional in the documentation generation but technically you need the underscores to generate the citations as links

Javierverbel: Ok, still is missing to reverse the [ABC]_ back to [ABC]

cryptographic_estimators/SDEstimator/sd_helper.py
Memphisd: @FloydZ could you please add the comment to the actual problematic line in the future, like this i always have to go to the file to check it 😇

cryptographic_estimators/SDEstimator/SDAlgorithms/stern.py
```diff
@@ -68,37 +71,38 @@ def l(self):
...
             9
         """
 
         return self._get_optimal_parameter("l")
 
     @optimal_parameter
     def p(self):
-        """
-        Return the optimal parameter $p$ used in the algorithm optimization
+        """Return the optimal parameter p used in the algorithm optimization.
```
Javierverbel, line 86: Sometimes there a new line after the ` """` sometimes no. Any reason for this?

cryptographic_estimators/SDEstimator/SDAlgorithms/ball_collision.py
```diff
@@ -66,80 +71,97 @@ def __init__(self, problem: SDProblem, **kwargs):
...
-        Return the optimal parameter $l$ used in the algorithm optimization
+        Return the optimal parameter $l$ used in the algorithm optimization.
 
-        EXAMPLES::
-
-            sage: from cryptographic_estimators.SDEstimator.SDAlgorithms import BallCollision
-            sage: from cryptographic_estimators.SDEstimator import SDProblem
-            sage: A = BallCollision(SDProblem(n=100,k=50,w=10))
-            sage: A.l()
+        Examples:
```
Memphisd, line 76: how does this napoleon extension for sphincs work? standard html doc generation via sphincs required capital spelling of EXAMPLES and TESTS and a double colon followed by a blank line? Maybe we should investigate this first to avoid later changes?

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -405,100 +382,119 @@ def initialize_parameter_ranges(self):
...
+                        }
                         if self._are_parameters_invalid(indices):
                             continue
                         yield indices
 
     def _time_and_memory_complexity(self, parameters: dict, verbose_information=None):
-        """
-        computes the expected runtime and memory consumption for the depth 3 version
-        """
+        """Computes the expected runtime and memory consumption for the depth 3 version."""
```
Javierverbel, line 497: Should we alway allow new line before and after the each ` """ `?
So `"""Computes the expected runtime and memory consumption for the depth 3 version."""` becomes
```
"""
Computes the expected runtime and memory consumption for the depth 3 version.
"""
```

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
Memphisd: yes maybe it is good to keep it, as for the later regex substitution a [xxx]_ might be more unique than [xxx]

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -30,29 +38,25 @@
...
     def __init__(self, problem: SDProblem, **kwargs):
-        """
-        Complexity estimate of BJMM algorithm in depth 2,3
+        """Complexity estimate of BJMM algorithm in depth 2,3.
 
-        The algorithm was introduced in [BJMM12]_  as an extension of [MMT11]_.
+        The algorithm was introduced in [BJMM12] as an extension of [MMT11].
 
-        expected weight distribution::
+        The expected weight distribution is:
```
Memphisd, line 45: In line with the above it should be "Expected weight distribution:"

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_dw.py
```diff
@@ -19,56 +19,56 @@
...
+
 filterwarnings("ignore", category=RuntimeWarning)
 
 
 class BJMMdw(SDAlgorithm):
     def __init__(self, problem: SDProblem, **kwargs):
-        """
-        Construct an instance of BJMM's estimator using *d*isjoint *w*eight distributions combined with
-         MitM-nearest neighbor search. [EB22]_, [MMT11]_, [BJMM12]_.
+        """Construct an instance of BJMM's estimator using disjoint weight distributions combined with MitM-nearest neighbor search. [EB22]_, [MMT11]_, [BJMM12]_.
```
Javierverbel, line 42: Here the underscores were not removed. @Dioprz  we have to be careful with this cases when adding the `_` back to those for which it was removed

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -30,29 +38,25 @@
...
 
-            +--------------------------+-------------------+-------------------+
-            | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |        p          |        p          |
-            +--------------------------+-------------------+-------------------+
+        +---------------------------+-------------------+-------------------+
```
Memphisd, line 47: the indentation should be kept (in line with the above at least) and for sphincs this ensured being displayed in a codeblock without spaces being removed

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_dw.py
```diff
@@ -220,10 +255,7 @@ def f(x): return log2(list_size) + 2 * \
         return l2_val
 
     def _time_and_memory_complexity(self, parameters: dict, verbose_information=None):
-        """
-        Computes the expected runtime and memory consumption for a given parameter set.
-        """
-
+        """Computes the expected runtime and memory consumption for a given parameter set."""
```
Javierverbel, line 258: New line before and after `"""`

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -64,7 +68,7 @@ def __init__(self, problem: SDProblem, **kwargs):
 
     def initialize_parameter_ranges(self):
         """
-        initialize the parameters p, l, p1 and for d=3 p2
+        Initialize the parameters p, l, p1, and for d=3, p2.
```
Memphisd, line 71: this docstring was wrong before already, it should say "Initialize parameter range for d"

cryptographic_estimators/SDEstimator/SDWorkfactorModels/bjmm.py
```diff
@@ -23,16 +23,14 @@
...
 
 class BJMMScipyModel(ScipyModel):
-    def __init__(self, par_names: list, problem: SDProblem, iterations: int, accuracy: int):
-        """
-        Optimization model for workfactor computation of BJMM algorithm in depth 3
-        """
+    def __init__(
+        self, par_names: list, problem: SDProblem, iterations: int, accuracy: int
+    ):
+        """Optimization model for workfactor computation of BJMM algorithm in depth 3."""
```
Javierverbel, line 29: New line before and after `"""`

cryptographic_estimators/SDEstimator/SDWorkfactorModels/prange.py
```diff
@@ -16,17 +16,14 @@
...
-from .workfactor_helper import representations_asymptotic, binomial_approximation, may_ozerov_near_neighbor_time
+from .workfactor_helper import binomial_approximation
 
 
 class PrangeScipyModel(ScipyModel):
     def __init__(self, par_names: list, problem: SDProblem, iterations, accuracy):
-        """
-        Optimization model for workfactor computation of Prange's algorithm
-        """
+        """Optimization model for workfactor computation of Prange's algorithm."""
```
Javierverbel, line 26: New line before and after `"""`

cryptographic_estimators/SDEstimator/SDWorkfactorModels/workfactor_helper.py
```diff
@@ -60,26 +60,20 @@ def binomial_approximation(n: float, k: float):
...
-    """
     return [(ru(x, y)) for _ in range(z)]
 
 
 def may_ozerov_near_neighbor_time(list_size: float, vector_length: float, target_weight: float):
-    """
-    computes the asymptotic runtime of the Nearest Neighbour Algorithm by
-    May-Ozerov [MO15]_
-    """
+    """Computes the asymptotic runtime of the Nearest Neighbour Algorithm by May-Ozerov [MO15]."""
```
Javierverbel, line 76: New line before and after `"""` and _

Javierverbel (changes requested):

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
```diff
@@ -169,49 +159,40 @@ def get_optimal_parameters_dict(self):
...
 
 
 class BJMMd2(SDAlgorithm):
     def __init__(self, problem: SDProblem, **kwargs):
-        """
-        Complexity estimate of BJMM algorithm in depth 2
+        """Complexity estimate of BJMM algorithm in depth 2.
 
-        The algorithm was introduced in [BJMM12]_  as an extension of [MMT11]_.
+        The algorithm was introduced in [BJMM12] as an extension of [MMT11].
```
Memphisd, line 170: reverting?

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
Memphisd: if I understood correctly there is an automatic html doc generation extension for this format (this napoleon sphincs extension) @Dioprz ?

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_pdw.py
```diff
@@ -33,39 +42,32 @@
...
+        Expected weight distribution:
 
-             +--------------------------+--------------------+--------------------+--------+--------+
-             | <-+ n - k - l1 - 2 l2 +->|<-+ (k + l1) / 2 +->|<-+ (k + l1) / 2 +->|   l2   |   l2   |
-             |       w - 2 p - 2 w2     |         p          |         p          |   w2   |   w2   |
-             +--------------------------+--------------------+--------------------+--------+--------+
+        +--------------------------+--------------------+--------------------+--------+--------+
+        | <-+ n - k - l1 - 2 l2 +->|<-+ (k + l1) / 2 +->|<-+ (k + l1) / 2 +->|   l2   |   l2   |
+        |       w - 2 p - 2 w2     |         p          |         p          |   w2   |   w2   |
+        +--------------------------+--------------------+--------------------+--------+--------+
```
Memphisd, line 53: in case we require the indentations, this has to be changed back (just marking in case)

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_plus.py
```diff
@@ -19,40 +19,45 @@
...
-            | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |        p          |        p          |
-            +--------------------------+-------------------+-------------------+
+        The expected weight distribution is as follows:
 
-        INPUT:
+        +--------------------------+-------------------+-------------------+
+        | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
+        |           w - 2p         |        p          |        p          |
+        +--------------------------+-------------------+-------------------+
```
Memphisd, line 51: .

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_plus.py
```diff
@@ -19,40 +19,45 @@
...
+        BJMM algorithm in terms of memory usage.
 
-        expected weight distribution::
+        For further reference, see [MMT11]_ and [BJMM12]_.
 
-            +--------------------------+-------------------+-------------------+
-            | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |        p          |        p          |
-            +--------------------------+-------------------+-------------------+
+        The expected weight distribution is as follows:
```
Memphisd, line 46: Expected weight distribution:

cryptographic_estimators/SDEstimator/SDAlgorithms/both_may.py
```diff
@@ -19,48 +19,49 @@
...
 
         For further reference see [BM18]_.
 
-            +-------------------+---------+-------------------+-------------------+
-            | <--+ n - k - l+-->|<-+ l +->|<----+ k / 2 +---->|<----+ k / 2 +---->|
-            |     w - w2 - 2p   |    w2   |         p         |         p         |
-            +-------------------+---------+-------------------+-------------------+
-
-        INPUT:
+        The structure is as follows:
```
Memphisd, line 42: Expected weight distribution:

cryptographic_estimators/SDEstimator/SDAlgorithms/both_may.py
```diff
@@ -19,48 +19,49 @@
...
 
-        - ``problem`` -- SDProblem object including all necessary parameters
+        +-------------------+---------+-------------------+-------------------+
+        | <--+ n - k - l+-->|<-+ l +->|<----+ k / 2 +---->|<----+ k / 2 +---->|
+        |     w - w2 - 2p   |    w2   |         p         |         p         |
+        +-------------------+---------+-------------------+-------------------+
```
Memphisd, line 47: .

cryptographic_estimators/SDEstimator/SDAlgorithms/dumer.py
```diff
@@ -29,39 +36,33 @@
...
 
-            +--------------------------+------------------+-------------------+
-            | <-----+ n - k - l +----->|<-- (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |       p          |        p          |
-            +--------------------------+------------------+-------------------+
-        INPUT:
+        +---------------------------+------------------+-------------------+
+        | <-----+ n - k - l +------>|<-- (k + l)/2 +-->|<--+ (k + l)/2 +-->|
+        |           w - 2p          |       p          |        p          |
+        +---------------------------+------------------+-------------------+
```
Memphisd, line 48: .

cryptographic_estimators/SDEstimator/SDAlgorithms/dumer.py
```diff
@@ -29,39 +36,33 @@
 class Dumer(SDAlgorithm):
     def __init__(self, problem: SDProblem, **kwargs):
         """
-        Complexity estimate of Dumer's ISD algorithm
+        Complexity estimate of Dumer's ISD algorithm.
 
         The algorithm was introduced in [Dum91]_.
 
-        expected weight distribution::
+        The expected weight distribution is as follows:
```
Memphisd, line 43: Expected weight distribution:

cryptographic_estimators/SDEstimator/SDAlgorithms/may_ozerov.py
```diff
@@ -30,26 +39,24 @@
...
         Introduced in [MO15]_.
 
-        expected weight distribution::
-
-            +--------------------------+-------------------+-------------------+
-            | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |        p          |        p          |
-            +--------------------------+-------------------+-------------------+
-        INPUT:
+        The expected weight distribution is as follows:
```
Memphisd, line 46: Expected weight distribution:

cryptographic_estimators/SDEstimator/SDAlgorithms/may_ozerov.py
```diff
@@ -30,26 +39,24 @@
...
 
-        - ``problem`` -- SDProblem object including all necessary parameters
+        +----------------------------+-------------------+-------------------+
+        | <-----+ n - k - l +------->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
+        |           w - 2p           |        p          |        p          |
+        +----------------------------+-------------------+-------------------+
```
Memphisd, line 51: .

cryptographic_estimators/SDEstimator/SDAlgorithms/may_ozerov.py
```diff
@@ -150,42 +148,36 @@ def get_optimal_parameters_dict(self):
...
     def __init__(self, problem: SDProblem, **kwargs):
-        """
-        Complexity estimate of May-Ozerov algorithm in depth 2.
+        """Complexity estimate of May-Ozerov algorithm in depth 2.
 
         [MO15] May, A., Ozerov, I.: On computing nearest neighbors with applications to decoding of binary linear codes.
         In: Annual International Conference on the Theory and Applications of Cryptographic Techniques. pp. 203-228 . Springer (2015)
 
-        expected weight distribution::
+        The expected weight distribution is as follows:
```
Memphisd, line 166: Expected weight distribution:

cryptographic_estimators/SDEstimator/SDAlgorithms/may_ozerov.py
```diff
@@ -150,42 +148,36 @@ def get_optimal_parameters_dict(self):
...
 
-            +--------------------------+-------------------+-------------------+
-            | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |        p          |        p          |
-            +--------------------------+-------------------+-------------------+
-        INPUT:
+        +--------------------------+-------------------+-------------------+
```
Memphisd, line 168: .

cryptographic_estimators/SDEstimator/SDAlgorithms/may_ozerov.py
```diff
@@ -293,73 +288,70 @@ def _time_and_memory_complexity(self, parameters: dict, verbose_information=None
...
     def __init__(self, problem: SDProblem, **kwargs):
-        """
-        Complexity estimate of May-Ozerov algorithm in depth 3.
+        """Complexity estimate of May-Ozerov algorithm in depth 3.
 
         [MO15] May, A., Ozerov, I.: On computing nearest neighbors with applications to decoding of binary linear codes.
         In: Annual International Conference on the Theory and Applications of Cryptographic Techniques. pp. 203-228 . Springer (2015)
 
-        expected weight distribution::
+        The expected weight distribution is as follows:
```
Memphisd, line 340: Expected weight distribution:

cryptographic_estimators/SDEstimator/SDAlgorithms/may_ozerov.py
```diff
@@ -293,73 +288,70 @@ def _time_and_memory_complexity(self, parameters: dict, verbose_information=None
...
 
-            +--------------------------+-------------------+-------------------+
-            | <-----+ n - k - l +----->|<--+ (k + l)/2 +-->|<--+ (k + l)/2 +-->|
-            |           w - 2p         |        p          |        p          |
-            +--------------------------+-------------------+-------------------+
-        INPUT:
+        +--------------------------+-------------------+-------------------+
```
Memphisd, line 342: .

cryptographic_estimators/SDEstimator/SDAlgorithms/prange.py
```diff
@@ -18,48 +18,51 @@
...
 from ..SDWorkfactorModels.prange import PrangeScipyModel
 
 
 class Prange(SDAlgorithm):
     def __init__(self, problem: SDProblem, **kwargs):
-        """
-        Construct an instance of Prange's estimator [Pra62]_
-
-        expected weight distribution::
+        """Construct an instance of Prange's estimator [Pra62].
```
Memphisd, line 34: _

cryptographic_estimators/SDEstimator/SDAlgorithms/prange.py
```diff
@@ -18,48 +18,51 @@
...
 
-            +--------------------------------+-------------------------------+
-            | <----------+ n - k +---------> | <----------+ k +------------> |
-            |                w               |              0                |
-            +--------------------------------+-------------------------------+
+        The expected weight distribution is as follows:
```
Memphisd, line 36: Expected weight distribution:

cryptographic_estimators/SDEstimator/SDAlgorithms/prange.py
```diff
@@ -18,48 +18,51 @@
...
 
-        INPUT:
+        +--------------------------------+-------------------------------+
+        | <----------+ n - k +---------> | <----------+ k +------------> |
+        |                w               |              0                |
+        +--------------------------------+-------------------------------+
```
Memphisd, line 41: .

cryptographic_estimators/SDEstimator/SDAlgorithms/stern.py
```diff
@@ -30,22 +38,20 @@ def __init__(self, problem: SDProblem, **kwargs):
...
         Construct an instance of Stern's estimator [Ste88]_, [BLP08]_.
 
-        Expected weight distribution::
+        Expected weight distribution:
 
-            +-------------------------+---------+-------------+-------------+
-            | <----+ n - k - l +----> |<-- l -->|<--+ k/2 +-->|<--+ k/2 +-->|
-            |          w - 2p         |    0    |      p      |      p      |
-            +-------------------------+---------+-------------+-------------+
+        +-------------------------+---------+-------------+-------------+
```
Memphisd, line 43: .

Memphisd, line 43: .

cryptographic_estimators/SDEstimator/SDWorkfactorModels/bjmm.py
```diff
@@ -70,37 +66,45 @@ def _build_model_and_set_constraints(self):
...
         return max(self.L1(x), self.L2(x), self.L3(x))
 
-    def _time_lists(self, x: float):
-        """
-        time to construct each list = max(size input list, size output list)
+    def _time_lists(self, x):
+        """Compute the time to construct each list.
+
+        The time to construct each list is the maximum of the size of the input
+        list and twice the size of the input list minus a constant.
```
Memphisd, line 83: Should be 
 The time to construct the lists is taken as the maximum of the involved lists sizes, i.e., the input and output lists.

cryptographic_estimators/SDEstimator/SDWorkfactorModels/workfactor_helper.py
```diff
@@ -93,20 +87,27 @@ def may_ozerov_near_neighbor_time(list_size: float, vector_length: float, target
...
+    )
 
 
 def representations_asymptotic(target_weight: float, weight_to_cancel: float, vector_length: float):
-    """
-    computes the asymptotic number of representations of a length-$vector_length$ weight-$target_weight$ vector
-    via the sum of two length-$vector_length$ weight-($target_weight$/2+$weight_to_cancel$) vectors
+    """Computes the asymptotic number of representations of a length-`vector_length` weight-`target_weight` vector.
+
+    This is done via the sum of two length-`vector_length` weight-(`target_weight`/2 + `weight_to_cancel`) vectors.
```
Memphisd, line 102: the meaning has changed. Please update to

Computes the asymptotic number of representations of a length-`vector_length` weight-`target_weight` vector constructed as sum of two length-`vector_length` weight-(`target_weight`/2 + `weight_to_cancel`) vectors.

cryptographic_estimators/SDEstimator/sd_helper.py
Memphisd: what floyd means is that citations have been removed. The citations [Bar07]_ [BLP08]_ should be mentioned. But the docstring was already inconsistent before

cryptographic_estimators/SDEstimator/sd_helper.py
```diff
@@ -208,26 +227,30 @@ def _mitm_nn_complexity(L: float, l: int, w: int, hmap: bool):
...
-    Complexity estimate of merging two lists exact
-    INPUT:
-    - ``L`` -- size of lists to be merged
-    - ``l`` -- amount of bits used for matching
-    - ``hmap`` -- indicates if hashmap is being used (Default 0: no hashmap)
-    EXAMPLES::
-        sage: from cryptographic_estimators.SDEstimator import _list_merge_async_complexity
-        sage: _list_merge_async_complexity(L1=2**16,L2=2**14,l=16,hmap=1) # random
-    """
+    """Compute the complexity of merging two lists exact.
```
Memphisd, line 230: Compute the complexity of merging two different sized lists on l bits.

Memphisd (changes requested):
Overall looks good. Just the AI changes of the docstrings sometimes lead to misunderstandings. We should be careful with reformulations of the explanations of the functions.

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
Memphisd: I assumed that this is part of the google styling guide. If not I also prefer the newlines. But I would also make it dependent on if there is any format requirement by the doc generation by sphincs

cryptographic_estimators/SDEstimator/sd_helper.py
FloydZ: I am always marking the problematic position, but admittedly, the message in this case is not particularly good. I think the problem is, that if someone applies changes to this file after I commented, the line numbers are misaligned, which leads to this misaligned references.
Is is possible to mark multiple lines in such a comment? That would make it a lot easier.

Dioprz:
Alright. Digging more in the Napoleon extension for Sphinx, I found that:

1. It's a preparser able to translate from Google-styled docstrings to typical reST python docstrings.
2. For this reason, we still have all the power of reST where needed.
3. With that said, I strongly discourage the usage of reST elements where avoidable. Basically because it's usage requires to learn about how to express a mathematical expression, a table, or whatever else. 
4. Napoleon provides a good number of sections and functionalities for most use-cases ([Supported sections](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections), [docstring types](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections) and [some configuration options](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections)). So let's just apply the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) here.
5. I completely agree with making the exception for hyperlink references with the trailing `_`.  I will apply the changes.
6. @FloydZ I was using backticks for `monospaced font` rendering. But looking at the [Sphinx documentation](https://github.com/numpy/numpydoc/blob/db19b2c45403128e83060b3b9901f00f7c1fc725/doc/format.rst#common-rest-concepts) on the topic, single backtick usage is context-dependant. What do you think we should do with that then?

Nice to hear your thoughts too: @Memphisd @Javierverbel

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
Dioprz: It's correct. Napoleon will produce the typical reST docstring as an intermediary step for a neat display.

cryptographic_estimators/SDEstimator/sd_helper.py
Dioprz: @FloydZ Yes, you can. Click the first line of the block, then press shift, and click the end line of the block.

Dioprz:
All the actionable comments were applied in [96c10fe](https://github.com/Crypto-TII/CryptographicEstimators/pull/176/commits/96c10fe2f052a05a664c870c8a561c9a92e71cde). My last question in the previous comment shouldn't be a blocker in case we want to merge this now; but I'm also open to new comments.

Please don't review #178 and #179 until I have implemented the same changes as here. I will ping you all when those are done.

Javierverbel:
I looks good to me @Dioprz . The let's ` MAYO-related` in the next PR docstring update, Maybe in #178

Javierverbel (dismissed):

Javierverbel (approved):

cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm.py
Memphisd: @Dioprz I don't understand why future editors have to fix from X to Y, X is what we want, or did you mean Y to X because of some automatic formatting that is applied?

Memphisd (approved):
This looks good to me. For future discussions, I’d appreciate it if we could include a less technical TL;DR section that summarizes the key points without too many complex Python terms. While it’s great to provide thorough references, it might be a bit optimistic to expect all reviewers to dive deeply into each one.
Note particularly to be applied in this thread, but often I find myself unable to handle PRs in time because I don't find the the time to study necessary references to follow up on the discussion related to it. It might also help to try to keep discussions as short as possible.

FloydZ (approved):

PR #180, opened 2024-09-19
Refactor/Makefile cleanup and improvements

Dioprz:
### Description

- `docker-run` now has dependency on `docker-build`. This is a sensible default as it saves one command, and if the build is not necessary, that extra step takes just one second.
- Each sensible and long command now uses local variables, making the code more clear and maintainable.
- The ad-hoc container creation and usage is improved in each command by using the `--rm` flag for `docker run`; avoiding the use of: 
  - `docker run ... && docker exec` (two commands instead of one). 
  - A post-execution container cleaning routine.
- Redefinition of the `docker-pytest` command, by splitting it in:
  - `docker-pytest-kat`
  - `docker-pytest-doctests`
- A new command (non functional yet, but functional after the migration) `docker-pytest-doctests-fast`.
- Removal of the unused `docker-pipeline` command.
- More meaningful prints.
- General improvements on naming and other minor things.

### Review process
-  Launch any command you want to test.

### Pre-approval checklist


Makefile
```diff
@@ -130,24 +128,18 @@ docker-testfast: docker-build
...
-	@echo "Cleaning container..."
-	@make stop-container-and-remove container_name="container-for-test"
 
-add-copyright:
-	@python3 scripts/create_copyright.py
-
-docker-pytest:
-	@echo "Removing previous container...."
-	@make stop-container-and-remove container_name="pytest-estimators" \
+docker-pytest-doctests: CONTAINER_NAME := "pytest-container"
```
FloydZ, line 136: very good idea! No need for sage in the doctests, which should significantly speed up the CI

FloydZ (dismissed):

Dioprz:
@FloydZ I needed multiple attempts to fix the CI. I removed the noisy commits, that's why the force-push. The only change was the removal of the flag `--it` in the `docker-pytest` commands.

And yes! We won't need more sage doctests once all the docstrings are migrated to pure python 🚀

FloydZ (dismissed):

Javierverbel (approved):

Javierverbel (dismissed):

Javierverbel (approved):

PR #181, opened 2024-09-19
Refactor/kat for PE estimator

Dioprz:
### Description
- Implementation of https://github.com/Crypto-TII/CryptographicEstimators/pull/171 for PEEstimator.

### [SOLVED] Help required
@Javierverbel @Memphisd @FloydZ 

Please take a look at this test: https://github.com/Crypto-TII/CryptographicEstimators/blob/3d052e447bb72f76fe7538c63da84d22c0f8e967/tests/external_estimators/PEEstimator/test_pe.sage#L55-L74

It is comparing not just one `actual_complexity` with the `expected_complexity`, but calculating two `actual_complexity` and taking the one that works the best.

This is not supported by the framework, so... Any idea?

I was thinking about identifying the edge cases inputs, and split the test in two, each one with a different `expected_complexity`.

Note: All the other tests were correctly migrated. The KAT generation is working (as you can notice from the serialized `.yaml`). And you can see where the problem is by looking at https://github.com/Crypto-TII/CryptographicEstimators/blob/3d052e447bb72f76fe7538c63da84d22c0f8e967/tests/internal_estimators/pe.py#L24-L37
So this is affecting the testing execution.

### Review process
```make docker-kat-tests```

### Pre-approval checklist

### CODE REVIEW

**Comment by Memphisd** [comment]
I looked into that. The issue is that there are some differences in the computation of the `w` parameter in the CryptographicEstimators versus the online available code. I think our version is fine.
In fact the optimal w we find is most of the time 1 _smaller_  than the online version (note that you should therefore use ` w=A.optimal_parameters()["w"] + 1 ` when executing our estimator the second time). However, to resolve the overall issue of two value testing I see two options:
1. Disregard the test completely: Could be arguable to do, as also there is a correction due to a mistake in the lee-brickell function in the online version 
2. Restrict to KAT test values that indeed have the difference of 1 (which is the case for most inputs) so that the only value to test against is  ```Leon(PEProblem(n, k, q), **leon_params).time_complexity(w=A.optimal_parameters()["w"] + 1 ) ```


In case we use 2. the condition to satisfy is ```(n,k,q)``` such that the code from Beullens:
```
def minimal_w(n,k,q):
    w = 1;
    S = 0
    while True:
        S += binomial(n,w)*(q-1)**(w-1)
        if S > 100*q**(n-k):
            return w, ceil(S/q**(n-k))
        w = w+1
```

returns an `w` that is one smaller than the optimal `w` of the Leon algorithm in the library.

**Comment by Dioprz** [comment]
Thank you for looking at this @Memphisd.

> Online version and implemented version are basically not the same anymore

I find this statement strong enough to find your option 1 as the better one. However, because 2. exists, I will wait for the opinion of @FloydZ and/or @Javierverbel to know what to do here.

**Comment by Javierverbel** [comment]
At this moment, I think it is better to remove them.

**Comment by FloydZ** [comment]
I think its fine to remove the test. And AFAIK we have a remark within the code, which states this problem. So whenever someone in the future encounter this discrepancy, he will at least find a note about this.

**Comment by FloydZ** [comment]
currently the tests fails with:
```
    def leon(input, epsilon=0.2):
        n, k, q = input

        # due to slightly different calculation of "number_of_weight_d_codewords" optimal w might differ by 1
        # for some edge cases
        actual_complexity_1 = Leon(PEProblem(n, k, q), **leon_params).time_complexity()
        actual_complexity_2 = Leon(PEProblem(n, k, q), **leon_params).time_complexity(
>           w=A.optimal_parameters()["w"] - 1
        )
E       NameError: name 'A' is not defined

tests/internal_estimators/pe.py:32: NameError
```

FloydZ (changes requested):
see my error msg

Dioprz:
@FloydZ I can't reproduce the error, and it looks like if you were using another branch or a non-updated docker container.

- Are you using `make docker-kat-tests`?

- Can you provide more details about the error?

FloydZ:
oh ok, nevermind, everything works now. No idea what happened

FloydZ (dismissed):

Javierverbel (dismissed):

Javierverbel (approved):

PR #200, opened 2024-10-14
Refactor/short doct tests

Javierverbel:
### Description
- Reduce the time taken to execute short doctests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ docker-doctests-fast: docker-build
 	    || true
 	@echo "Running short doctests..."
 	@docker run --name ${CONTAINER_NAME} --rm -it ${IMAGE_NAME} sh -c "\
-          pytest --skip-long-doctests  --doctest-modules -n auto -vv -s \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -168,7 +168,7 @@ docker-doctests-fast: docker-build
 	    || true
 	@echo "Running short doctests..."
 	@docker run --name ${CONTAINER_NAME} --rm -it ${IMAGE_NAME} sh -c "\
-          pytest --skip-long-doctests  --doctest-modules -n auto -vv -s \
+          pytest --skip-long-doctests  --doctest-modules -n auto -vv \
```
FloydZ, line 171: @Dioprz  wrote at one point, that the `-s` flag does something important. But I dont remember what.

FloydZ (approved):

Makefile
Dioprz: Yes, it shows the prints produced by the tests (so it's not captured by pytest). However, it is only needed in the KAT-tests command, so this is okey!

FloydZ (approved):

PR #201, opened 2024-10-15
Refactor/python migration cleanup

Dioprz:
### Description

This PR is the last step in the python migration milestone for the library. It includes the next aspects:

1. A general cleanup in the library, removing all the remanent Sage imports.

2. Deprecation of the old `setup.py` and the introduction of the new python standard, `pyproject.toml` as the file to declare all the metainformation necessary to build the library.

2. A refactor in the Dockerfile removing the necessity for hardcoding the library dependencies. It preserves the performance on build/rebuild for the `docker`-commands.

4. A great refactor and cleanup of the Makefile, as well as the introduction of local commands equivalent to the docker-based ones (`make doctests`, `make doctests-fast`, `make tests-all`, `make functional-tests`, `make kat-tests` )

### Review process
- For the `pyproject.toml`
  - Install the library:
     ```shell
     python3 -m venv .venv
     source .venv/bin/activate
     pip install .
     ```
  - Look at the accuracy of the information declared in the file.
- For the `Dockerfile` and the `Makefile`
  - Test any command you want (and it would be great if you can test the docker and local version of the commands: `make [docker-]doctests`, `make [docker-]doctests-fast`, `make [docker-]tests-all`, `make [docker-]functional-tests`, `make [docker-]kat-tests` .

### Pre-approval checklist


Javierverbel (dismissed):

cryptographic_estimators/estimation_renderer.py
```diff
@@ -20,7 +20,7 @@
 from .helper import concat_all_tables, round_or_truncate
 from copy import deepcopy
 from prettytable import PrettyTable
-from sage.all import RR
+from numpy import float64
```
FloydZ, line 23: why numpy? I just grepped over the repo and thats the only place where we include numpy

FloydZ (approved):
If run all tests (docker and non-docker) on my x86 machine in the past 4h or so and they worked. Great work @Dioprz !

cryptographic_estimators/estimation_renderer.py
Dioprz: @FloydZ Happy to hear that!! :rocket: And thank you!

About numpy, yes, that's tricky. The reason is because we need to replace the check for the rows, from being in sage RR, to being of the type `float` or `float64`, as seen here: 
https://github.com/Crypto-TII/CryptographicEstimators/blob/94d574f404f4defdb5a62fa34c971657ce97637a/cryptographic_estimators/estimation_renderer.py#L124-L126

Otherwise, the rounding of the rows doesn't work. The reason is probably because of previous calculations using SciPy or SymPy, as both of them use numpy under the hood (if I remember correctly).

Javierverbel (approved):

PR #217, opened 2024-12-08
Fix/hashimoto only for mq underdefined

Javierverbel:
### Description
- Make Hashimoto algorithm to be only executed for underdefined MQ instances


ricardovp85 (approved):
It looks Ok to me.

fmerino21 (approved):

PR #243, opened 2025-02-12
Feat/fuzzing

FloydZ:
### Description
This PR contains
- the possibility to run fuzzing tests
- fixes for found issues with the fuzzer

### Check
 For example
``` bash
make docker-fuzzing-run FUZZ_ARGS="--sd"
make docker-fuzzing-run FUZZ_ARGS="--mq --bytes 2"
make docker-fuzzing-run FUZZ_ARGS="--ranksd --bytes 2 -max_total_time=300"
```


Dioprz:
(I changed this to a draft to avoid accidental merging)

After some reads about Atheris and fuzzy testing, this sounds very interesting. Happy to learn about its existence.

Let me know if I can help with something here, @FloydZ.

FloydZ:
Hi @Dioprz ,
currently I'm not really 100% happy with the solution. There are two main issues:
- `atheris` only allows for a single function to be fuzzed at each time. Thats why I have code like:
```python
#atheris.Setup(sys.argv, SDFuzz)
...
atheris.Setup(sys.argv, RankSDFuzz)
...
#atheris.Setup(sys.argv, MAYOFuzz)
```
And you need enable/disable the estimator by hand. If you know how to circumvent that, without adding a bash script, which calls the python code multiple times. let me know.
The second problem is the exception handling. At first I thought catching `ValueError` is enough to separate between "real problematic" errors in the code, and errors which are already caught by the class constructor. But turns out errors like `log2(0)` are also `ValueErrors`, hence I have this ugly:
```python
if type(e) != ValueError or str(e) == "math domain error":
```
check to difference between those two. If you have a better solution for this, I would really appreciate it.

FloydZ:
Apparently, sonar cloud thinks that:
``` python
if w <= 0 or k <= 0:
```
always evaluates to false, for what ever reason.

Dioprz:
> Apparently, sonar cloud thinks that:
> 
> ```python
> if w <= 0 or k <= 0:
> ```
> 
> always evaluates to false, for what ever reason.

The full code snippet failing (well, one of them) is this one:

```python
    def __init__(self, n: int, k: int, w: int, **kwargs):
        super().__init__(**kwargs)
        if k <= 0 or n <= 0 or w <= 0:
            raise ValueError("n, k, w must be positive integers")
        if k > n:
            raise ValueError("k must be smaller or equal to n")
        if w > n - k:
            raise ValueError("w must be smaller or equal to n-k")
        if w <= 0 or k <= 0:
            raise ValueError("w and k must be at least 1")
```

And looking at it... I think it makes sense because you will never be able to reach the last `if` branch, as the same conditions will trigger the first `if` branch. Makes sense, or I'm overlooking something?

Javierverbel (dismissed):

FloydZ:
updated PR. Can be reviewed

Javierverbel (dismissed):

Memphisd:
I think this atheris library should be preinstalled in the dev-docker. It seems rather elaborate to install it, on Mac at least.

Javierverbel (changes requested):
Second command doesn't work for me. I get
```
Unable to find image 'cf-fuzzer:latest' locally
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "./tests/fuzzing/fuzz.py": permission denied: unknown.
```

FloydZ:
updated the PR

Issue #233, opened 2025-01-21
Add Parameters into Table Output


FloydZ:
If you optimize over a lot of values you often get output like this:
```
: 27 m: 26 o: 2 k: 14
+----------------------+--------------+------------------------------------------------------------+
|                      |              |                          estimate                          |
+----------------------+--------------+-------+--------+-------------------------------------------+
| algorithm            | attack_type  |  time | memory |                 parameters                |
+----------------------+--------------+-------+--------+-------------------------------------------+
| DirectAttack         |   forgery    |  55.1 |   19.9 | {'k': 2, 'a': 12, 'variant': 'Hashimoto'} |
| KipnisShamir         | key-recovery | 110.5 |   12.5 |                     {}                    |
| ReconciliationAttack | key-recovery |  77.7 |   24.6 |      {'k': 5, 'variant': 'las_vegas'}     |
| IntersectionAttack   | key-recovery | 138.7 |   20.6 |      {'k': 0, 'variant': 'las_vegas'}     |
| ClawFinding          |   forgery    |  65.4 |   55.4 |         {'X': 55.565, 'Y': 48.435}        |
+----------------------+--------------+-------+--------+-------------------------------------------+
n: 27 m: 26 o: 3 k: 10
+----------------------+--------------+------------------------------------------------------------+
|                      |              |                          estimate                          |
+----------------------+--------------+-------+--------+-------------------------------------------+
| algorithm            | attack_type  |  time | memory |                 parameters                |
+----------------------+--------------+-------+--------+-------------------------------------------+
| DirectAttack         |   forgery    |  55.1 |   19.9 | {'k': 2, 'a': 12, 'variant': 'Hashimoto'} |
| KipnisShamir         | key-recovery | 102.5 |   13.1 |                     {}                    |
| ReconciliationAttack | key-recovery |  73.7 |   24.6 |      {'k': 4, 'variant': 'las_vegas'}     |
| IntersectionAttack   | key-recovery | 126.7 |   20.6 |      {'k': 0, 'variant': 'las_vegas'}     |
| ClawFinding          |   forgery    |  65.4 |   55.4 |         {'X': 55.565, 'Y': 48.435}        |
+----------------------+--------------+-------+--------+-------------------------------------------+
```
which makes it sometimes hard to figure out which table belongs to which parameter set. Maybe we can add a flag `parameters_inside=true` to the code: 
```
E = MAYOEstimator(n=n, m=m, o=o, k=k, q=q)
[...]

PR #244, opened 2025-02-12
Feat/added the possibility to show the parameters within the table

FloydZ:
### Description
- this PR adds the possibility to show the problem parameters within the table. Implements https://github.com/Crypto-TII/CryptographicEstimators/issues/233

Test script:
```python 
from cryptographic_estimators.SDEstimator import SDEstimator
SD = SDEstimator(n=15, k=10, w=5)
SD.table(parameters_inside=True)

```


cryptographic_estimators/estimation_renderer.py
```diff
@@ -78,7 +79,11 @@ def _create_initial_table_containing_algorithm_column(self, estimation: dict) ->
         """
         table = PrettyTable([BASE_ALGORITHM])
         table.padding_width = 1
-        table.title = ' '
+        if self._problem is not None:
+            table.title = ", ".join([str(a) for a in self._problem.get_parameters()])
```
Memphisd, line 83: is it possible to add the parameter name like (n=100, k=50, w=10) rather than (100, 50, 10)

Memphisd (changes requested):
Please use in future branches the branch naming convention, which is lower-case snake_case

cryptographic_estimators/estimation_renderer.py
FloydZ: I dont know how. There is no data struct in python which holds this information. The only thing we have (in case of SD) is `sd_constants.py`:
```python
SD_CODE_LENGTH = "code length"
SD_CODE_DIMENSION = "code dimension"
SD_ERROR_WEIGHT = "error weight"
```
I mean I could somehow use reflections to parse the names of the parameters from the construction of the problem class. Buts thats like super hacky.

FloydZ:
> I was wondering why this has to be implemented for every estimator separately.
I was hoping not. Maybe we could use the `args` and `kwargs` arguments. So every `table()` function  within each estimator gets those two arguments. And only the `table` function in `base_estimator` gets the actual `parameters_inside` argument.

FloydZ:
addressed issues by andre. ready for review

Javierverbel (approved):
Looks nice. Should we add one or two doctests for this feature?

PR #252, opened 2025-02-25
bug/fix docker file

FloydZ:


Memphisd:
isn't sage needed to generate the kats?

Dioprz:
> isn't sage needed to generate the kats?

@Memphisd Yes, it is

@FloydZ, I will send a commit in some minutes with a fix for this

Dioprz:
Some advance there. The build now is working, but the testing process is failing. I'm looking on this.

Also, FTR:

My commit changes were quite drastic, as I completely changed the Dockerfile (even the base image). The error we were receiving was coming from the fact that Sage was using the system level libraries instead of the `pip` installed ones, so even if `pip` installed `setuptools X.Y.Z`, if the system had the version `A.B.C`, it will take the `A.B.C` one from the system.

In normal situations this should be fixable by just changing the base image *version*, but Sage was removed from the official repository of Ubuntu, so if you upgrade to a newer version, you also need to install it from scratch. Because the process looked painful, I decided to go directly for the official [Sage image](https://hub.docker.com/r/sagemath/sagemath), and start fixing things from there. Let's see how that works.

Dioprz:
Alright, this should be fixed and ready for review.

Sage's docker image is much more aggressive with the environment variables, and it basically hijacks everything (when you call python3 from the container, you are actually calling the python3 of sage, for example). To solve it, I needed to change the CI-related docker commands in our Makefile to add a sage prefix so everything works as expected.

.github/workflows/build-and-test.yml
```diff
@@ -35,4 +35,8 @@ jobs:
           rm -rf /tmp/.buildx-cache
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
       - name: Run functional, doctests and KAT tests
-        run: make docker-tests-all
+        run: |
```
FloydZ, line 38: please remove commented commands

Makefile
```diff
@@ -92,37 +92,37 @@ docker-doctests: docker-build
 	@make stop-container-and-remove container_name=${CONTAINER_NAME} \
 		|| true
 	@echo "Running doctests..."
-	@docker run --name ${CONTAINER_NAME} --rm ${IMAGE_NAME} sh -c "${DOCTESTS_COMMAND}"
+	docker run --name ${CONTAINER_NAME} --rm ${IMAGE_NAME} sage -python3 -m${DOCTESTS_COMMAND}
```
FloydZ, line 95: missing whitespace behind `-m `.

FloydZ:
@Dioprz found two little issues. Plus the CI still errors, but not in the actual tests. The sonar cloud has some problems:
```bash
  sed -i "s/<source>\/home\/cryptographic_estimators\/cryptographic_estimators<\/source>/<source>\/github\/workspace\/cryptographic_estimators<\/source>/g" coverage.xml
```
It seems to be that it tries to change something in a path, where it already cd into. Maybe the problem is the:
```
WORKDIR "/home/sage/cryptographic_estimators/"
```
in the docker file.
So my guess is to change line 51 in `run-pytest-and-sonarcloud.yml` to:
```
        run: sed -i "s/cryptographic_estimators<\/source>/<source>\/cryptographic_estimators<\/source>/g" coverage.xml
```

Dioprz:
Solved at https://github.com/Crypto-TII/CryptographicEstimators/pull/257. Closing PR.

Issue #250, opened 2025-02-21
Warnings in doctests


Javierverbel:
I get the following warnings when executing `make doctests`
```
cryptographic_estimators/SDEstimator/sd_estimator.py::cryptographic_estimators.SDEstimator.sd_estimator.SDEstimator.table
CryptographicEstimators/cryptographic_estimators/SDEstimator/SDAlgorithms/bjmm_dw.py:248: RuntimeWarning: The iteration is not making good progress, as measured by the 
    improvement from the last five Jacobian evaluations.
    l2_val = int(fsolve(f, 50)[0])

cryptographic_estimators/SDEstimator/sd_estimator.py: 5379956 warnings
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    return fsolve(lambda x: v - (-x * log2(x) - (1 - x) * log2(1 - x)), 0.0000001)[0]
```

Dioprz:
Fixed by https://github.com/Crypto-TII/CryptographicEstimators/pull/253 (merged at https://github.com/Crypto-TII/CryptographicEstimators/pull/257).

PR #253, opened 2025-02-25
Fix/warnings during doctest

Memphisd:
### Description
Addresses issue #250 
- Switch from fsolve to root function to avoid warning 
- access array element in lambda function via x[0]


Dioprz:
Changes included at https://github.com/Crypto-TII/CryptographicEstimators/pull/257. Closing PR.

PR #269, opened 2025-05-22
Version Function

FloydZ:
### Description
- this PR adds a `.version()` function to the `BaseEstimator` class which should help identifying the exact version a user is using if its unclear.

### Review process
```python
from cryptographic_estimators.SDEstimator import SDEstimator

SDE = SDEstimator(n=100, k=50, w=10, include_quantum=True)
SDE.version()
```
this code should print something like:
```
Version: 2.0.0
```

Please test this function in different scenarios, like if the test file is not in the repo. If the repo is not installed, etc...


dio-rz (commented):
I can't provide quick testing now because I need to set up a Nix dev shell to install the package locally; but for now I would ask you to please change the docstrings to follow the [Google style docstrings](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html) ([or from our docs](https://github.com/Crypto-TII/CryptographicEstimators/blob/main/docs/github/docstrings_and_html.md)).

FloydZ:
redone the whole thing. Its based on `pyproject.toml`

Issue #279, opened 2025-07-23
Integrated wedge attack to UOV like schemes by Ran


Javierverbel:
https://eprint.iacr.org/2025/1143

PR #282, opened 2025-11-18
Feat: wedge attack on UOV and MAYO

Javierverbel:
### Description
- Wedge attack estimator in UOV and MAYO [(Issue)](https://github.com/Crypto-TII/CryptographicEstimators/issues/279)


FloydZ (approved):
The PR looks fine. The only question I have is, why is the Wedge Attach not part of the MQEstimator? Because now you implemented twice basically the same in two different files.

PR #309, opened 2026-06-23
Fix/regsd solution density

tl2cents:
### Description
  - Account for the expected number of RegSD solutions when computing repeated-trial time (`q_iter`) for RegularISDPerm,
  RegularISDEnum, and RegularISDRep.
  - The solution density is taken from `problem.nsolutions`.

  - Check that the three regular ISD algorithms use it for their total time estimate.
  - Verify small testing examples.

  ### Pre-approval checklist


FloydZ (approved):

FloydZ:
thanks, good catch.

PR #312, opened 2026-08-31
Add just guess algorithm

Javierverbel:
Adding the JustGuess algorithm (EUROCRYPT2026)


FloydZ (dismissed):

FloydZ:
I bypass the merging rules, because only sonar is failing.

